PR #1, opened 2018-05-29
Add first draft of CLI

killercup:
Given a git URL, it'll

- ask for project name
- clone the repo
- replace `{{project-name}}`, `{{crate_name}}`, and `{{authors}}`
  placeholder
- init a new repo

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,25 @@
+# EditorConfig helps developers define and maintain consistent
+# coding styles between different editors and IDEs
+# editorconfig.org
+
+root = true
+
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+
+[*.rs]
+indent_style = space
+indent_size = 4
+
+[*.toml]
+indent_style = space
+indent_size = 4
+
+[*.md]
+trim_trailing_whitespace = false
--- a/src/main.rs
+++ b/src/main.rs
@@ -46,12 +46,12 @@ main!(|args: Cli| {
 
     let engine = liquid::ParserBuilder::new().build();
     let mut placeholders = liquid::Object::new();
-    placeholders.insert("project-name".to_owned(), liquid::Value::scalar(&name));
+    placeholders.insert(String::from("project-name"), liquid::Value::scalar(&name));
     placeholders.insert(
-        "crate_name".to_owned(),
+        String::from("crate_name"),
         liquid::Value::scalar(&ident_case::RenameRule::SnakeCase.apply_to_field(&name)),
     );
-    placeholders.insert("authors".to_owned(), liquid::Value::scalar(&get_authors()?));
+    placeholders.insert(String::from("authors"), liquid::Value::scalar(&get_authors()?));
 
     let progress = indicatif::ProgressBar::new_spinner();
     progress.tick();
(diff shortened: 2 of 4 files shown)


killercup:
Oh hey @ashleygwilliams I think we kinda forgot to talk about this in Paris since I started doing other stuff? This should be somewhat good-to-go as a first draft

.editorconfig
```diff
@@ -0,0 +1,25 @@
+# EditorConfig helps developers define and maintain consistent
```
ashleygwilliams: just curious- why this and not just a cargo fmt test? (i'm open to it, but i dont want style battles ;))

src/main.rs
```diff
@@ -0,0 +1,162 @@
...
+
+    let _template = GitRepository::clone(&args.git, &project_dir)
+        .with_context(|_e| format!("Couldn't clone `{}`", &args.git))?;
+
+    fs::remove_dir_all(&project_dir.join(".git"))
+        .context("Error cleaning up cloned template")?;
+
+    let engine = liquid::ParserBuilder::new().build();
+    let mut placeholders = liquid::Object::new();
+    placeholders.insert("project-name".to_owned(), liquid::Value::scalar(&name));
```
ashleygwilliams: would really prefer these `to_owned`s be `String::from`s or `to_string`s for readability and contributors sake

ashleygwilliams: unless this is somehow required for liquid?? 🤔

src/main.rs
```diff
@@ -0,0 +1,162 @@
...
+}
+
+main!(|args: Cli| {
+    let name = match &args.name {
+        Some(ref n) => n.to_string(),
+        None => query_name()?,
+    };
+
+    let project_dir = env::current_dir()
+        .unwrap_or_else(|_e| ".".into())
```
ashleygwilliams: curious about this `into` here

.editorconfig
killercup: This is so editors like vim and vscode can (with plugins) automatically set the correct settings. I usually add it to my projects. Want me to delete it?

I'll also run rustfmt!

src/main.rs
killercup: I use `into` when I don't know/care what the type is supposed to be, but I know how to get it. I think it's OsString here? Or maybe PathBuf? I can be  more explicit if you like

src/main.rs
killercup: Changed to `String::from`.

I'm super lax with these type conversions, sorry. But you are right and I agree, for other people reading the code, `.to_string()` gives you way more information. This is useful when learning -- you can see that what you are inserting in this line is a `String` and a `liquid::Value`.

But: I don't think you really need to know that. I trust the compiler will tell me if the types don't work, and otherwise I don't really care about if it's a `String` or a `&str` or a `Path(Buf)` or an `OsStr(ing)` or a `Cow<str>` or a `Rc<str>`or whatever. Because… Why would I? This doesn't change the content and it's an API I don't control! Out of curiosity, I hovered over the `placeholders` binding and RLS told me (and [the docs](https://docs.rs/liquid/0.14.3/liquid/type.Object.html) agree): This is a `HashMap<String, Value>` behind a type alias. I had no idea -- and I didn't need. Why am I trying to explain this? I'm not entirely sure. It's something I've been thinking of a lot, how explicit code needs to be.

(In this specific case, I'm also pretty sure I copy-pasted this from the liquid docs. I would've written `into()` because it's faster to type.)

killercup:
ping @ashleygwilliams (just remembered this is still open, no rush)

ashleygwilliams:
@killercup lol goddamnit, sorry! i'm gonna merge this and we just handle nits and such in other issues/PRs.

PR #20, opened 2018-07-11
Killercup integration tests

ashleygwilliams:


killercup:
TIL this is necessary on Windows! (Will need to document that somewhere for people who try to write CLI apps and might want to add that to quicli)

PR #25, opened 2018-07-16
refactor ALL THE THINGS

### LINKED ISSUE #23: refactor/modularize
in prep for publishing this and opening to contributors, let's break the src and tests up into modules

ashleygwilliams:
fixes #23

### DIFF
--- a/src/git.rs
+++ b/src/git.rs
@@ -1,7 +1,7 @@
 use git2::{build::CheckoutBuilder, build::RepoBuilder, Repository as GitRepository,
            RepositoryInitOptions};
 use quicli::prelude::*;
-use std::fs::remove_dir_all;
+use remove_dir_all::remove_dir_all;
 use std::path::PathBuf;
 use Cli;
 
--- a/src/git.rs
+++ b/src/git.rs
@@ -1,0 +1,24 @@
+use git2::{build::CheckoutBuilder, build::RepoBuilder, Repository as GitRepository,
+           RepositoryInitOptions};
+use quicli::prelude::*;
+use std::fs::remove_dir_all;
+use std::path::PathBuf;
+use Cli;
+
+pub fn create(project_dir: &PathBuf, args: Cli) -> Result<GitRepository> {
+    Ok(RepoBuilder::new()
+        .bare(false)
+        .with_checkout(CheckoutBuilder::new())
+        .clone(&args.git, &project_dir)?)
+}
+
+pub fn remove_history(project_dir: &PathBuf) -> Result<()> {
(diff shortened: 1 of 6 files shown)


src/git.rs
```diff
@@ -0,0 +1,24 @@
+use git2::{build::CheckoutBuilder, build::RepoBuilder, Repository as GitRepository,
+           RepositoryInitOptions};
+use quicli::prelude::*;
+use std::fs::remove_dir_all;
```
killercup: wait why are we no longer using the remove_dir_all crate?

src/git.rs
```diff
@@ -0,0 +1,24 @@
...
+use std::path::PathBuf;
+use Cli;
+
+pub fn create(project_dir: &PathBuf, args: Cli) -> Result<GitRepository> {
```
killercup: why not take `&Path` and `&Cli`?

killercup: also I'd call this `create_from_git`

src/progressbar.rs
```diff
@@ -0,0 +1,5 @@
+use indicatif::ProgressBar;
```
killercup: how do you feel about `extern crate`s that are not in the root module? I would probably put the `extern crate indicatif;` here

src/main.rs
```diff
@@ -62,131 +60,18 @@ main!(|args: Cli| {
...
+    git::create(&project_dir, args)?;
+    git::remove_history(&project_dir)?;
 
-    for entry in WalkDir::new(&project_dir) {
-        let entry = entry?;
-        if entry.metadata()?.is_dir() {
-            continue;
-        }
+    let mut template = template::new();
+    template = template::substitute(&name, template)?;
```
killercup: For what is currently happening here the `template::new()` is confusing -- it is a (constant) constructor of an empty object that is immediately passed to `substitute`. And `substitute` is only ever called with this (still constant) value.

I'd change this whole thing to `let template = template::substitute(&name);`

src/git.rs
ashleygwilliams: oh crap. mistakes were made. thanks for catching!

src/git.rs
ashleygwilliams: re Cli, i just didn't need to. but don't really care. the module is called git, so it reads `git::create` - i'd rather not have the redundancy

src/progressbar.rs
ashleygwilliams: not generally a huge fan :/

src/main.rs
ashleygwilliams: wfm

PR #32, opened 2018-07-17
v0.1.0

ashleygwilliams:


killercup:
If this fails on appveyor i'll wear a "fails CI on windows" shirt for a week

PR #35, opened 2018-07-18
refactor(tests): split helpers out into mods

### LINKED ISSUE #33: split test helpers into own module

ashleygwilliams:
fixes #33 

"i like small files and i cannot lie"


tests/integration/helpers/mod.rs
```diff
@@ -0,0 +1,2 @@
+pub mod project;
+pub mod project_builder;
```
killercup: How about you `pub use project_builder::dir` (and maybe some other types required to make rustc happy) instead of exposing both mods?

ashleygwilliams:
lol. feel free to open a PR with a preferred solution :P

PR #56, opened 2018-07-26
fix crate_name

ashleygwilliams:
fixes #41 and #34

PR #62, opened 2018-07-27
Move from ident_case to heck

csmoe:
Closes https://github.com/ashleygwilliams/cargo-generate/issues/57


killercup:
I'm so sorry @csmoe I just noticed that I was _way_ to quick in accepting this! While this fixes the one apparent usage of ident_case, @ashleygwilliams pointed #63 out to me (based on #56). In addition to that, I've also missed that with this patch we don't store the kebab-case version any longer, though!

That's all on me -- we'll follow up some of these issues in new PRs (cf. #64). Just wanted to let you know and thank you again for taking initiative here!

Issue #65, opened 2018-07-27
warn user if we change given name's case


ashleygwilliams:
right now, we receive a name from the user, force it to kebab case, and make the project. we should check the user's entered name and warn them if we are changing the case of the name they passed.

epage:
Other possible approaches
- Templates have validation code in a [hook](https://github.com/ashleygwilliams/cargo-generate/issues/18) that errors on an invalid name
- Ask for a human friendly name and then ask for the package name, offering a default that is a `heck` slug of the human friendly name.

k0pernicus:
@ashleygwilliams Maybe I can work on that issue - if you are agree (and no one is already working on this one :))

ashleygwilliams:
go for it @k0pernicus ! :)

ashleygwilliams:
closed by #84

PR #84, opened 2018-08-06
Issue/65

k0pernicus:
Add a first solution to issue #65.
Instead of computing each time the kebab/snake case string, I stored those strings in the struct.

Also, I implemented a method which verify if the original project name and the kebab case name is different, or the same.

Finally, I added a warning message inside `main.rs` in order to warn the user if the project will be renamed or not (using the previous struct method).

### DIFF
--- a/src/projectname.rs
+++ b/src/projectname.rs
@@ -4,20 +4,24 @@ use heck::{KebabCase, SnakeCase};
 /// for handling casing.
 pub struct ProjectName {
     pub user_input: String,
-    pub kebab_case_name: String,
-    pub snake_case_name: String,
 }
 
 impl ProjectName {
     pub fn new(name: &str) -> ProjectName {
         ProjectName {
             user_input: name.to_string(),
-            kebab_case_name: name.to_kebab_case(),
-            snake_case_name: name.to_snake_case(),
         }
     }
 
-    pub fn force_renaming(&self) -> bool {
-        self.user_input != self.kebab_case_name
+    pub fn kebab_case(&self) -> String {
+        self.user_input.to_kebab_case()
+    }
+
+    pub fn snake_case(&self) -> String {
+        self.user_input.to_snake_case()
+    }
+
+    pub fn is_crate_name(&self) -> bool {
+        self.user_input == self.kebab_case()
     }
 }
--- a/src/projectname.rs
+++ b/src/projectname.rs
@@ -17,11 +17,5 @@ impl ProjectName {
         }
     }
 
-    pub fn kebab_case(&self) -> String {
-        self.user_input.to_kebab_case()
-    }
-
-    pub fn snake_case(&self) -> String {
-        self.user_input.to_snake_case()
     }
 }
--- a/src/projectname.rs
+++ b/src/projectname.rs
@@ -17,5 +17,7 @@ impl ProjectName {
         }
     }
 
+    pub fn force_renaming(&self) -> bool {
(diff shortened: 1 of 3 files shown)


src/projectname.rs
```diff
@@ -3,21 +3,21 @@ use heck::{KebabCase, SnakeCase};
 /// Stores user inputted name and provides convenience methods
 /// for handling casing.
 pub struct ProjectName {
-    user_input: String,
+    pub user_input: String,
```
k0pernicus: As we don't use `user_input` outside the struct, I can remove the `pub` here.

k0pernicus:
So, it seems the Travis build fails due to this error:
```
$ cargo fmt -- --write-mode=diff
Unrecognized option: 'write-mode'
```

Is this error can be related to [this cargo issue](https://github.com/rust-lang-nursery/rustfmt/issues/2279)?

src/projectname.rs
```diff
@@ -3,21 +3,21 @@ use heck::{KebabCase, SnakeCase};
...
         ProjectName {
             user_input: name.to_string(),
+            kebab_case_name: name.to_kebab_case(),
+            snake_case_name: name.to_snake_case(),
         }
     }
 
-    pub fn kebab_case(&self) -> String {
-        self.user_input.to_kebab_case()
-    }
```
killercup: Super tiny nit: I'd keep these methods as getters instead of using the field names (and make all fields private), so it's easier to change this again in the future.

What you could however do is give these semantically meaningful names, like `as_crate_name()` and `as_identifier()`.

killercup:
> Unrecognized option: 'write-mode'

Oh, it seems the new rustfmt version is now on stable. I think we need to use `--check` or something like this. (I'm quite annoyed this broke.)

ashleygwilliams:
yeah i will update the cargo fmt check right now, sorry about that! i'll give this a review soon after :)

src/projectname.rs
k0pernicus: Thanks for this review :)
The `crate_name` is kebab-case right?

k0pernicus: Is the nit about getters instead of using field names is part of another Rust tips than it will be easier to change? (just for my culture)

killercup: Yeah, crate names are usually written in kebab-case (snake_case is also allowed).

Using getters is something I personally tend to do quite often, partially also because I also like to add random debug statements and breakpoints (which is very easy to add to a function but practically impossible to add to struct field accesses). The official API guidelines have a recommendation in their [future proofing](https://rust-lang-nursery.github.io/api-guidelines/future-proofing.html#structs-have-private-fields-c-struct-private) chapters to not use public fields because changing them is a breaking change to your crate's API. (This is of course less of an issue for a binary crate.)

k0pernicus: Great to hear!
Thanks :)

k0pernicus:
I still have an error with Travis.  
This error is related to `rustfmt`.

The `rustfmt` associated to my local version format the line 101 as ```
let dir_name = if force { name.raw() } else { name.kebab_case() };
```.  
However, the formatting related to the Heroku instance is ```
let dir_name =
    if force { name.raw() } else { name.kebab_case() };
```.

Is there an option to set in my local configuration?
Like maximum characters per line, something like that...?

Thanks a lot!

k0pernicus:
@ashleygwilliams No problem - I would be pleased to help you in solving other issues!
I will check open issues.

Issue #81, opened 2018-08-02
split into shallow main and throw functionality in lib


ashleygwilliams:
there is a universe (#80) where  i would like to use this tool as a library. it's a pretty common pattern for cli apps to have a shallow main that primarily uses a lib. we're real close to that, but we need a lib.rs to get it over the hump.

DD5HT:
I've a working Prototype with a shallow main. But I still have to improve the error handling in the library before I can push it. I've also refactored some of the logical blocks into separate functions.
`src/bin/client.rs`
```rust
#[macro_use]
extern crate quicli;
extern crate cargo_generate;

use cargo_generate::{Cli, generate};
use quicli::prelude::StructOpt;

main!(|_cli: Cli| {
    generate(_cli);
});
```

ashleygwilliams:
cool! if you want to start a WIP PR (work-in-progress PR) i'm happy to work with you on it- we may end up needing to take quicli out... i think that's fine. also we can have a top level ? in main :D - that might help! (it's a new feature so i'm not sure if you knew about it)

Issue #79, opened 2018-08-02
improve error message when passed a git repo that does not exist


dineshadepu:
I am getting an error while creating a cli-wg project `ggrs`.

```
dinesh@Personal /Users/dinesh $  
|  Mac => cargo generate --git https://github.com/rust-clique/cargo-template-cli
🤷  Project Name: grrs
🔧   Creating project called `grrs`...
error: authentication required but no callback set; class=Net (12)
```

killercup:
Yep, that template doesn't exist yet, see 
https://github.com/rust-lang-nursery/cli-wg/issues/53 (and I should really add that TODO to the tutorial!)

We should improve the error message here. I think we can at least add some context (what repo, more info about the [git error type](https://docs.rs/git2/0.7.4/git2/struct.Error.html)), and also print out some common errors (like missing repos or missing permissions).

ashleygwilliams:
go for it @rexto - probably want to build off of what @DD5HT is working on in #85! (would totally accept a PR that's branched off his work)

DD5HT:
I think it would be really good if you write a function like this in 
`git.rs `:
```rust
    pub fn pretty_git_err(err: Error) -> String {
        YOUR IMPLEMENTATION
    }
```

taurr:
Nowadays we actually get

`Error: ⛔   Git Error: unexpected http status code: 404; class=Http (34)`

(404 = Not found, which is correct for a bogus URL)
or

`Error: ⛔   Git Error: unexpected http status code: 400; class=Http (34)`

(400 = Bad Request, which can also be correct as in this case I used the <username>/<repo> syntax, and github found the user, but not the repo)

PR #85, opened 2018-08-07
WIP: Shallow main

DD5HT:
#81 I'm working on building a little generate lib.
I try to handle all errors internally so it will just print out the problem to the user without the use of  `?`.
To make the main really shallow, we may have to remove `quicli`. 

Also #79 is handled a little better in this build since the tells the user that it's a git specific error.
 
I still have to do some more refactoring and error handling. I'm also thinking about adding a few more specific tests to check the error handling. 

I'm also open for more suggestions eg. better names and co. 🐬

### DIFF
--- a/src/bin/client.rs
+++ b/src/bin/client.rs
@@ -1,0 +1,35 @@
+#[macro_use]
+extern crate quicli;
+extern crate cargo_generate;
+
+use cargo_generate::{Cli, generate};
+use quicli::prelude::StructOpt;
+
+/// Generate a new Cargo project from a given template
+///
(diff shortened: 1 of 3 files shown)


DD5HT:
Okay it kinda works now, my questions are now if there is anything I should change? 
Should I handle some of the more unlikelier errors with println! too? or maybe use an early exit macro?

ashleygwilliams:
glad to have you back @DD5HT ! hope the move went well. this is absolutely still wanted- thanks so much! feel free to ask any questions!

Issue #71, opened 2018-07-29
Support repository banches


deg4uss3r:
I set my templates up in one repository and was branching for differences in my templates (see [cli-templates](https://github.com/deg4uss3r/cli-template) for my rough drafts); however, as of now `cargo-generate` only supports main branch templating. 

To me it makes sense to have all my custom templates in one repostory but have the ability to call them by branch, vice several different repositories. Also it might make sense for offical templates to have branches with different specializations or layouts.

Related to RFC #47

ashleygwilliams:
@deg4uss3r this is a good call and shouldn't be much overhead over what we are doing using git anyways! thanks for filing :)

deg4uss3r:
Thanks @ashleygwilliams! If you'd like I could try to help with this issue :)

ashleygwilliams:
that'd be great @deg4uss3r ! feel free to file a WIP PR to start, and always happy to help with questions if you have any :)

ashleygwilliams:
closed by #94

PR #87, opened 2018-08-07
Adding branches to template repository pulls

deg4uss3r:
This is a work in progress (WIP) PR for Issue #71  

PS- Sorry was on holiday :)

### DIFF
--- a/src/git.rs
+++ b/src/git.rs
@@ -9,6 +9,7 @@ use Args;
 pub fn create(project_dir: &PathBuf, args: Args) -> Result<GitRepository> {
     Ok(RepoBuilder::new()
         .bare(false)
+        .branch(&args.branch)
         .with_checkout(CheckoutBuilder::new())
         .clone(&args.git, &project_dir)?)
 }
--- a/src/interactive.rs
+++ b/src/interactive.rs
@@ -1,6 +1,7 @@
 use console::style;
 use dialoguer::Input;
 use emoji;
+use heck::KebabCase;
 use quicli::prelude::Error;
 use regex;
 
@@ -13,6 +14,13 @@ pub fn name() -> Result<String, Error> {
             style("Project Name").bold()
         )).interact()?;
         if valid_ident.is_match(&name) {
+            println!(
+                "{} {} `{}`{}",
+                emoji::WRENCH,
+                style("Creating project called").bold(),
+                style(&name.to_kebab_case()).bold().yellow(),
+                style("...").bold()
+            );
             break name;
         } else {
             eprintln!(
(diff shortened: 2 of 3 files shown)


ashleygwilliams:
no sorry necessary! hope you had a nice holiday :)

deg4uss3r:
Hey! Almost done here, but I couldn't reproduce the non-zero exit code from `cargo fmt --all -- --check` locally with the same call or with `rustfmt --check`.

DD5HT:
@deg4uss3r did you read [this](https://github.com/ashleygwilliams/cargo-generate/blob/master/CONTRIBUTING.md)? you are missing a new line at the end of `main.rs`
these are my versions which run totally fine:
```
rustfmt 0.8.2-stable (08da30d 2018-06-06)
rustc 1.28.0 (9634041f0 2018-07-30)
```

deg4uss3r:
@DD5HT Thanks, didn't see that in the [contribute](https://github.com/ashleygwilliams/cargo-generate/blob/master/CONTRIBUTING.md) documentation, but fixed.

deg4uss3r:
Let me know if there's any tests or documentation you'd like me to add and I would be happy to work on that :)

posborne:
I didn't see this PR at first and put together some changes for this feature as it is something I ended up needing this evening.  Here's the patchset I put together: https://github.com/ashleygwilliams/cargo-generate/compare/master...posborne:add-git-branch-support?expand=1

I think the main issue with the approach in this PR is that having a hard-coded default of "master" rather than omitting the checkout parameter will break some repos that might have a different default branch on the remote than "master" (on some repos, this may be something like "develop").

Let me know if I should open up a PR with my changes (changes for tests got a bit gross, admittedly).

deg4uss3r:
Hey!

I’m on mobile so haven’t had a proper look at the changes, but do you think for 0.2.0 it would be sufficient to make branch a required option?

ashleygwilliams:
hey sorry for the delay @deg4uss3r - i was on vacation and am slowly catching up on all the things. i think this looks great!

that being said i do agree with @posborne that the hardcoded default branch will probably cause some issue. @posborne do you think you could rebase your solution on top of this one? i'd like to merge this and have yours be an improvement upon it- if that's possible! lemme know if you think that rebase would be too difficult to be worth it :)

posborne:
@ashleygwilliams I can probably make a rebase work if that is the preferred approach.  I would need to open a separate PR with that.  I think my changes would replace these except for the extra print which doesn't really seem directly related to the change (https://github.com/ashleygwilliams/cargo-generate/pull/87/files#diff-137d81403bc89c5575b8f43e343d07ffR17) -- is that something we want to include with this changeset?

ashleygwilliams:
closing in favor of https://github.com/ashleygwilliams/cargo-generate/pull/94#pullrequestreview-160316324 - sorry for the extremely bad communication. if there's something else you'd like to do on the project @deg4uss3r please let me know and it is yours!

Issue #82, opened 2018-08-03
Do some cleanup on the template


TeXitoi:
Hi!

I would need to remove some files from the template.

Usecase: migrate https://github.com/TeXitoi/blue-pill-quickstart to cargo-generate

Some files are here only for the purpose of explaining the repository. I'd like these files to be deleted when using `cargo generate`. In this case, the files to remove are `LICENSE` `README.md` and `st-link-v2-blue-pill.jpg`.

DD5HT:
So you want something like .gitignore for cargo generate.  #48 a config file could also solve this problem.

ashleygwilliams:
closed by #96

PR #91, opened 2018-08-09
added ignore files capabilites 

DD5HT:
Solving #82 

Maybe a little hacky to solve this problem but it can read a `.genignore `file and remove the contents specified in this file.
I would also make it optional with the default = true

Do we really want to call it `.genignore` or are there better names for it?

> TODO:
> + Boundary Checks against malicious ignore files necessary?


ashleygwilliams:
closing in favor of https://github.com/ashleygwilliams/cargo-generate/pull/96

PR #94, opened 2018-09-19
Add support for specifying git branch

posborne:
This mimics the `--branch` option available when performing
a `cargo install` from git.

I created this without first noticing https://github.com/ashleygwilliams/cargo-generate/pull/87 but after some discussion here it seemed like it made sense to open this as an alternative that addresses issues related to assuming that "master" was always the default branch on the remote.

CC @deg4uss3r


posborne:
ping @ashleygwilliams (because I too get lost in a sea of Github notifications sometimes).

PR #96, opened 2018-10-01
Dd5 ht genignore

ashleygwilliams:
rebase of #91

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,4 +9,5 @@ matrix:
     script:
     - cargo test --all --locked
     - rustup component add rustfmt-preview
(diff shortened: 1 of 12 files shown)


DD5HT:
Hello Ashley :) I'm glad that you're active again and I'm also looking forward  helping you in the development of this tool  :+1:  Since it took a little longer I have to check the code again and look what I wanted to do :) I think we can just remove the sanitize function in this case. The Idea was that a user could write a malicious genignore file which would remove ~ for example. But this should not be possible in this implementation anyway.

ashleygwilliams:
also i can add a test that makes sure that the tool does the correct thing when passed something like `~`

PR #98, opened 2018-10-01
Safety check

### LINKED ISSUE #97: test that unsafe ignore operations are not possible
follow up to #96

ashleygwilliams:
fixes #97 

this is a pretty blunt test. curious what you think @DD5HT


ashleygwilliams:
merging so i can turn my attention to #85 but comments here are very welcome and will def be addressed before release

PR #108, opened 2018-10-04
feat(ignore): always ignore genignore

### LINKED ISSUE #102: clean up .genignore automagically
the current impl of `.genignore` requires that you explicitly add the `.genignore` file in order for it to be cleaned up. as @killercup pointed out when i spoke to him, this is pretty surprising. because the `.genignore` file should always be cleaned up, let's automagically do so, and not require users to list it in the `.genignore` file itself.

ashleygwilliams:
fixes #102

### DIFF
--- a/src/ignoreme.rs
+++ b/src/ignoreme.rs
@@ -3,18 +3,17 @@ use remove_dir_all::*;
 use std::collections::HashSet;
 use std::ffi::OsStr;
 use std::fs::remove_file;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 ///takes the directory path and removes the files/directories specified in the
 /// `.genignore` file
 /// It handles all errors internally
 pub fn remove_uneeded_files(dir: &PathBuf) {
-    if let Some(items) = get_ignored(dir) {
-        remove_dir_files(items);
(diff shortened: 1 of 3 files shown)


ashleygwilliams:
lol too bad it didn't work tho. gonna sort this today as it's blocking the release that i'm a week late on now :(

src/ignoreme.rs
```diff
@@ -36,11 +36,7 @@ fn get_ignored(location: &PathBuf) -> Option<Vec<PathBuf>> {
     for x in all_set.difference(&ign_set) {
         output.push(x.to_owned());
     }
-    if output.is_empty() {
-        None
-    } else {
-        Some(output)
-    }
+    Some(output)
```
k0pernicus: I checked the file and didn't found any clue about an Option type other that `Some(output)`.
If I did not made any mistake in reading the code, why do we return an `Option` type if it's always `Some(output)`... ?

ashleygwilliams: yeah i think that's a good call, previously it returned an option because there may be nothing in Output, i'll change it

k0pernicus:
Ok to me :ok_hand:

PR #115, opened 2018-10-13
new shallow main 

DD5HT:
I did create a new Branch and deleted the old PR, rebasing didn't really work out that well in my case 😅 
Now I'm gonna refactor the lib.rs and then it is ready for review😁


ashleygwilliams:
would still be really interested in this work @DD5HT ! let me know if you still want to finish it <3

DD5HT:
@ashleygwilliams I just did a rebase.
 I'm going to do some better error handling and clippy runs the next few days. 
Then I'm finally done and can work on other Issues 🐬 🐬

src/lib.rs
```diff
@@ -108,38 +128,50 @@ main!(|_cli: Cli| {
...
+        None
+    } else {
+        Some(project_dir)
+    }
+}
 
-    let template = template::substitute(&name, force)?;
+fn progress(name: &ProjectName, dir: &PathBuf, force: bool) {
+    let template =
+        template::substitute(name, force).expect("Error: Can't substitute the given name.");
```
ashleygwilliams: i think we'd like to still use the `?` mark operator- is there a reason you replaced it with `expect`?

src/lib.rs
```diff
@@ -108,38 +128,50 @@ main!(|_cli: Cli| {
...
 
     let pbar = progressbar::new();
     pbar.tick();
 
-    template::walk_dir(&project_dir, template, pbar)?;
+    template::walk_dir(dir, template, pbar).expect("Error: Can't walk the directory");
```
ashleygwilliams: same here

DD5HT:
I added some better error handling changed some `expect()/unwrap()` to `?`  I think if the tests run through and you are happy with the changes you could merge it. @ashleygwilliams

ashleygwilliams:
i'm a big fan of human panic @DD5HT ! would def be excite to include it :)

PR #119, opened 2018-10-27
Use cargo to pull git remotes.

ChristopherMacGown:
Since cargo-generate doesn't set a credentials callback for
git2::Repository, it can't pull authenticated repos. Cargo provides an
[authentication callback](https://github.com/rust-lang/cargo/blob/6a7672ef5344c1bb570610f2574250fbee932355/src/cargo/sources/git/utils.rs#L409-L617) in `sources::git::utils::with_authentication`,
but the callback isn't public.

Rather than copying the callback to this repo, it was more maintainable
long-term to just let Cargo handle pulling git-repos and managing
submodules.

I also added a test for submodules and found a bug in the template
processing where submodules didn't have their git metadata excluded from
the template processing. So now the template processing excludes git
metadata no matter where it's located.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1,4 +1,9 @@
 [[package]]
+name = "adler32"
+version = "1.0.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "aho-corasick"
 version = "0.6.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


ChristopherMacGown:
**BEFORE**
```chris@cmacgown:~ ∵ cargo generate --git https://github.com/rustwasm/wasm-pack-template.git --name test
🔧   Creating project called `test`...
error: authentication required but no callback set; class=Ssh (23)
chris@cmacgown:~ ∵ 
```
**AFTER**
```chris@cmacgown:~ ∵ ./target/debug/cargo-generate generate --git https://github.com/rustwasm/wasm-pack-template.git --name test
🔧   Creating project called `test`...
The given paths are neither files nor directories! "/Users/chris/test/.genignore"
✨   Done! New project created /Users/chris/test
chris@cmacgown:~ ∵ 
```

Issue #117, opened 2018-10-25
PascalCase project name variable


passcod:
I started writing a template for https://github.com/passcod/notify/tree/next backends, but there's no PascalCase substitution? I want to do this:

```rust
const BACKEND_NAME: &'static str = "{{project-name}}";

pub struct {{ProjectName}} {
    //
}

impl Backend for {{ProjectName}} {
    fn name() -> &'static str {
        BACKEND_NAME
    }

    // ...
}

// etc
```

ashleygwilliams:
sorry for the delay in respones! i would absolutely accept a PR to add this.

passcod:
That would be ideal. Don't have lots of time right now but I might look (if
ashley is okay with this approach) during yuletide or thereabouts.

ashleygwilliams:
it was literally forever, but this finally landed. cutting a 0.3.0 release now.

Issue #48, opened 2018-07-25
RFC: Have a config file for cargo-generate?


killercup:
## Context

When cargo-generate tries to generate a project, it may require meta-information, e.g. which placeholders to look for, or how to find files to apply substitutions to. While currently, this is defined by hard-coded rules in the tool itself, in the future this might be configurable.

## Questions

Should templates be able to contain (optional) config files to pass parameters and meta information to cargo-generate?

## Possibles solutions

- Yes, and it'll be in TOML format with a schema like

    ```toml
    [template]
    name = "foo"
    repository = "https://github.com/rust-lang-nursery/cargo-template-foo"
    placeholders = ["project-name", "author"]
    ```

- Yes, and it can also be a subsection of a `Cargo.toml` file
- Yes, and it can also be a subsection of a `book.toml` file
- No

anirudhb:
Have the template itself contain a `Parameters.toml`, and then cargo-generate could read from it and prompt the user?

epage:
RE Format

Ordering is important and ideally we don't rely on parser-specific implementation details to get ordering.  From the toml spec:
> Key/value pairs within tables are not guaranteed to be in any specific order.

Do we need `type` fields?  Personally, I hope string and [choice](https://cookiecutter.readthedocs.io/en/latest/advanced/choice_variables.html) good enough.

Values should be templates that are evaluated before the next question. 

For more possibly-related thoughts, see https://github.com/ashleygwilliams/cargo-generate/issues/17

PR #140, opened 2018-12-28
feat: Support case filters in templates

epage:
This adds
- capitalize (stdlib)
- pascal_case (custom)
- kebab_case (custom)
- snake_case (custom)

Some might be duplicate of built-in behavior but that is in prep for
experimenting with #48.

We could add more case filters but they seemed unlikely to be used.

There are a lot of other string filters in liquid that could be useful
as well as a "slugify" (turn arbitrary strings into identifiers) or
"deunicode" which converts unicode characters to ascii.  For now, I'm
starting conservatively.

Fixee #117

### DIFF
--- a/src/template.rs
+++ b/src/template.rs
@@ -1,6 +1,7 @@
 use cargo;
 use console::style;
 use emoji;
+use heck::{CamelCase, KebabCase, SnakeCase};
 use indicatif::ProgressBar;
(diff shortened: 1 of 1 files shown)


ashleygwilliams:
closing in favor of https://github.com/ashleygwilliams/cargo-generate/pull/149

PR #146, opened 2019-03-29
Fix windows release on appveyor

jaysonsantos:
This should build a release version only when it is building a tag, thus
fixing the problem where it does not back the exe final in the final
.tar.gz file.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -14,6 +14,7 @@ test_script:
 
 before_deploy:
   - ps: |
+        cargo build --release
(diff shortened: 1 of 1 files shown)


.appveyor.yml
```diff
@@ -14,6 +14,7 @@ test_script:
 
 before_deploy:
   - ps: |
+        cargo build --release
```
ashleygwilliams: we can actually update the test_script above (which will run a build) to include `--release`:

`cargo test --release --all --locked`

ashleygwilliams: hrm. i may be mistaken about this. looking into it.

ashleygwilliams: i was mistaken. let's go with your patch for now!

PR #157, opened 2019-04-25
Fix assert! macro usage

rasendubi:
There is a bug in rustc that allows adding invalid trailing tokens to the `assert!` macro call. They are currently ignored but are going to produce errors in the future.

Fix assert! macro usage to remove extra tokens.

For more information, see https://github.com/rust-lang/rust/issues/60024 and https://github.com/rust-lang/rust/pull/60039


ashleygwilliams:
failing on cargo-generate. ill merge and then take care of the formating

ashleygwilliams:
failing on cargo-generate. ill merge and then take care of the formatting

ashleygwilliams:
failing on cargo-generate. ill merge and then take care of the formatting

PR #161, opened 2019-05-30
Added new feature: filename can contains template tag(s)

k0pernicus:
This PR permits to automatically rename a file that contains template tags, as requested here: https://github.com/ashleygwilliams/cargo-generate/issues/159.


DD5HT:
looks good

k0pernicus:
@ashleygwilliams I am working on it right now :)

k0pernicus:
@ashleygwilliams I added two new doc lines about this feature.  
Please let me know if want to modify something, or add more docs/examples :)

ashleygwilliams:
this looks great. @xortive how does this affect your pr #174 ? i'd like to get this into the release if it's not too disruptive.

xortive:
@ashleygwilliams A filename would have to get past the include / exclude logic (so if operating in whitelist mode, they must explicitly include a glob matching the pre-replace filename) with how I implemented my PR, but It is possible to separate that from the filename replace logic to handle the case where you may want to template a file but _not_ replace it's contents by performing a separate directory walk just for looking at file paths.

steveklabnik:
This vaguely reminds me of a recent discussion around C++ modules, that is:

```cpp
#define flib flob
module flib;
```

defines a module named `flob`. Tricky!

I do think that 

> However, should the include / exclude logic (and subsequent template directory walk that goes through matched files and replaces their contents) operate before or after the filenames being processed?

Is the big question here. I think I agree with @xortive here:

> IMO, separate filename processing, with the include / exclude logic (file contents processing) running before, makes the most sense and allows for the most options.

Keeping them separate, so that you don't have complex interplay between rules, and processing them before, so that by the time you get to contents, it's only dealing with contents, seems good. I think it's also easier to try this way first, and if it turns out to be bad for some reason, combine them, then it is to start with them combined, and then pull it apart.

src/template.rs
```diff
@@ -141,27 +141,41 @@ pub fn walk_dir(
...
-                style(filename.display()).bold()
+                style(filepath.display()).bold()
+            )
+        })?;
+
+        // Check if the filename does not contains any
+        // template
+        let filename = filepath.to_str().unwrap();
+        let parsed_filename = engine.clone().parse(filename)?.render(&template)?;
+        let _ = fs::rename(&filepath, Path::new(&parsed_filename)).with_context(|_e| {
```
steveklabnik: what's the reason for `let _ =` here? You're already `?`-ing the `Result` away, right?

xortive:
```suggestion
        fs::rename(&filepath, Path::new(&parsed_filename)).with_context(|_e| {
```

xortive: this smells of lingering clippy lint fixes that don't make sense anymore, so I'm going to just fix this to get this PR merged in!

Issue #163, opened 2019-06-28
Add username placeholder


Sreyas-Sreelal:
There's already an authors placeholder for filling up authors field in Cargo.toml.But there could be situations when there's only the need of username and not in `username <email>` format.Since the code already fetches username for authors field,the same can be adjusted for fetching both.

I did make a working solution [here](https://github.com/Sreyas-Sreelal/cargo-generate/commit/f58f88e5d07eb5daa0283463f98389d6ca0a99a2).But being not sure whether it's the most ideal solution and thought it'd be wise to open an issue and discuss with people who are more experienced before opening the pull request.

Issue #17, opened 2018-07-10
Custom variables.


epage:
For my  python projects, I've found cookiecutter's custom variables really helpful.  I'm assuming it would be the same for `cargo-generate` (including unexpected cases like #14).

Ideas from cookiecutter
- variables are an ordered set of keys/values. 
  - Keys are variable names
  - Values are defaults.  The values are [templates that are rendered using all the previous values](https://cookiecutter.readthedocs.io/en/latest/advanced/templates_in_context.html).
  - Variables can also be [choices](https://cookiecutter.readthedocs.io/en/latest/advanced/choice_variables.html)
- Users can [globally configure variables](https://cookiecutter.readthedocs.io/en/latest/advanced/user_config.html), removing the need to be prompted for them.
- [User input is cached off for replay](https://cookiecutter.readthedocs.io/en/latest/advanced/replay.html)
  - I've not needed this as much but its useful for template debugging.

I'm also assuming that environment-derived variables (author) would still be gathered but built-in policy like "name" would not be.
- In place of this, there could be an option for variables to be set on the commandline (`cargo generate --variable name some-crate`).
- For built-in magic like case changing, can be provided as liquid filters to the template author. 
- Unsure of a good way to handle ident verification.

theduke:
As posted on Reddit, my suggestion for variables:

`cargo-generate.toml ` in stand-alone repos and possibly in Cargo.toml for crates that are templates or crates that contain templates.
```toml
variables = [
  { type = "bool", name = "use_serde", prompt = "Add serde support?", default = true },
  { type = "choice", name = "db_type", prompt = "Which database would you like to use?", options = [ "mysql", "sqlite", "postgres"] },
 ...
]
```

`lib.rs`
```
{% if use_serde %} ... {% else %} ... {% endif %}

{% if db_type = "mysql% } .....
```

japaric:
This is a use case from the embedded WG. I'm filling it here since it's related to the requested feature but feel free to move into its own issue if that makes sense.

To use the [cortex-m-quickstart] "template" we have the user `cargo-clone` the crate and then fill in some values manually. This last part is easy to get wrong so we have an extensive [troubleshooting] section to help the user fix things if they entered a wrong value. It would be great to turn this template into a cargo-generate template and have cargo-template request and validate user input when creating the template; that way we wouldn't need such a long troubleshooting guide!

[...]

PR #164, opened 2019-06-29
add username placeholder

Sreyas-Sreelal:
Implements #163


src/authors.rs
```diff
@@ -6,7 +6,7 @@ use std::env;
 /// Taken from cargo and thus (c) 2018 Cargo Developers
 ///
 /// cf. https://github.com/rust-lang/cargo/blob/d33c65cbd9d6f7ba1e18b2cdb85fea5a09973d3b/src/cargo/ops/cargo_new.rs#L595-L645
-pub fn get_authors() -> Result<String, failure::Error> {
+pub fn get_authors() -> Result<(String, String), failure::Error> {
```
ashleygwilliams: i think we'd want to return a Map of some type so that it's more discoverable as to what these two strings are

ashleygwilliams: additionally, this edits code that we've taken directly from Cargo. for maintenance purposes it might be nicer to have the username logic live outside of this and treat this code a bit like a vendored dependency.

Sreyas-Sreelal:
On second thought,how about prompting user to enter username? If the user didn't enter anything,fetch username from system configurations.There could be situations when an user want to enter their github username instead of OS username

Sreyas-Sreelal:
@ashleygwilliams?

k0pernicus:
Hi @Sreyas-Sreelal, I am very sorry for the delay...
Are you still working on this issue?

MarcoIeni:
> On second thought,how about prompting user to enter username? If the user didn't enter anything,fetch username from system configurations.There could be situations when an user want to enter their github username instead of OS username

Personal opinion: I think that this variable should just fetch username from system configuration.

Additional variables to be prompted to the user should be defined as custom variables. See #17. But Again, just my opinion based on personal taste.

sassman:
closing by inactivity. Also I think we have enough flexibility for template authors to define custom fields right now.

PR #165, opened 2019-07-05
chore: Update to liquid 0.19

epage:
This will resolve a borrow checker warning.


epage:
For background, the intention of the new filter API is:
- Reduce error checking in filters (the new API does type checking, number of args checking, etc)
- Speed up parsing by front loading parameter handling to parse time rather than render time
- Allow auto-generating help by inspecting what filters are included in the parser

The last one is the one you all will care about the most though it still needs some additional APIs to get it to work.
 
What we're still not happy with (ie potential future breakages)
- Even with the derive macros, writing the new API is a bit clunky to write
- The organization of the filters that `liquid` exposes.  We're still trying to figure out what is the best way to expose std and non-std filters.

PR #169, opened 2019-07-24
Used vendored openssl for macosx builds

xortive:
Travis CI's OSX environment causes the build to dynamically link to openssl@1.1 which isn't installed by default on mac machines. For consistency with the linux binary, we can use the vendored-openssl feature to sidestep this issue, which will eliminate having to install openssl@1.1 for users consuming the release binary. This increases the build size for macos by ~2mb.


ashleygwilliams:
i am very happy to merge this so long as we feel confident that the binary produced improves the experience (i expect it will but am not sure!)

PR #174, opened 2019-08-06
Feature: Add file for configuring include / exclude of files to template

### LINKED ISSUE #43: RFC: Only replace variables in files with specific extension?
## Context

Currently, cargo-generate tries to find and replace all `{{ blocks }}` in all files. This is the most trivial implementation, but not necessarily the best or most user friendly one.

As written in <https://github.com/ashleygwilliams/cargo-generate/issues/14#issuecomment-405957200> and <https://github.com/ashleygwilliams/cargo-generate/issues/40#issuecomment-406536791>, an alternative might be to instead only consider files whose names match a specific pattern as templates.

## Questions

- Should we filter by file extensions?
- If yes, which ones?
- If yes, where to define this extension? (Also cf. #48)

## Possibles solutions

- Yes, with `.template`

    (Because it's pretty obvious and also not bound to any implementation. It has the added advantage that projects that already have `.hbs` or `.liquid` files are unlikely to conflict.)

- Yes, defined in config file (cf. #48), defaulting to `.tplt`

- No

## Notes

The implementation should be pretty straightforward. I assume it's a matter of expanding [this condition](https://github.com/ashleygwilliams/cargo-generate/blob/9850761ca2895f38a96d73e945ee476e590181f1/src/template.rs#L34-L36) and adding a test or two.

### LINKED ISSUE #48: RFC: Have a config file for cargo-generate?

### LINKED ISSUE #124: generating a project with tera template
If the template project includes [tera] (https://github.com/Keats/tera) html templates cargo-generate fails with:
```
error:  Error replacing placeholders `/root/rustlang/test/templates/index.html`
caused by: liquid: Invalid index
  with:
    index=appname
```
It seems to me generate should just ignore all placeholders except those it knows about. Maybe issue a warning,but definitely not fail on them.

### LINKED ISSUE #144: Cannot generate project with binary files
Hi, I'm using cargo-generate on a project with binary files in it, got this error. Even list the file in .genigore does not seem to help.

This is the error msg.
```
🔧   Creating project called `flutter-app`...
error: liquid: Cannot read file
cause: stream did not contain valid UTF-8
  with:
    path=/Users/bach/Develop/flutter-app/rust/assets/icon.icns
```

icns is macos icon resource.

xortive:
Adds support for a file which allows for templates to configure which files should be processed, either using a whitelist method (include), or a blacklist method (exclude), with include being preferred over exclude if both are present. This mirrors the behavior used by cargo for the configuration options of the same name in `Cargo.toml`, and uses some of the same code.

closes #144
closes #124
closes #48
closes #43
closes https://github.com/cloudflare/worker-emscripten-template/issues/1

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,0 +1,188 @@
+use failure::{bail, Fail};
+use serde::{Deserialize, Serialize};
+use std::fs::File;
+use std::io::Read;
+use std::path::Path;
+use globset::GlobSet;
+use toml;
+
+#[derive(Deserialize, Debug, PartialEq)]
+pub struct Config {
+    pub template: Template,
+}
+
+struct Template {
+    pub name: Option<String>,
+    pub repository: Option<String>,
+    pub placeholders: Option<String>,
+    pub parse_matcher: Option<GlobSet>,
+}
+
+#[serde(try_from="TemplateRaw")]
+impl TryFrom<TemplateRaw> for Template {
+    fn from(raw_template_config: TemplateRaw) -> Result<Self, failure::Error> {
+        match raw_template_config {
+            TemplateRaw {
+                include: Some(_),
+                exclude: None,
+                ..
+            } => {
+
+
+            },
+            TemplateRaw {
+                include: None,
+                exclude: Some(_),
+                ..
+            } => {
+
+
+            },
+            TemplateRaw {
+                include: Some(_),
+                exclude: Some(_),
+                ..
+            } => {
+                let err = TemplateConfigError::XorFieldError {
+                    field_a: "include".into(),
+                    field_b: "exclude".into(),
+                };
+
+                Err(err.into())
+            },
+        }
+    }
+}
+
+#[derive(Deserialize, Debug, PartialEq)]
+struct TemplateRaw {
+    pub name: Option<String>,
+    #[serde(rename = "repo")]
+    pub repository: Option<String>,
+    pub placeholders: Option<Vec<String>>,
+    pub include: Option<Vec<String>>,
+    pub exclude: Option<Vec<String>>,
+}
+
+#[derive(Debug, Fail)]
+enum TemplateConfigError {
+    #[fail(
+        display = "config fields {} and {} are mutually exclusive",
+        field_a, field_b
+    )]
+    XorFieldError { field_a: String, field_b: String },
+}
+
+impl Config {
+    pub fn new(path: &Path) -> Result<Self, failure::Error> {
+        let mut file = File::open(path)?;
+        let mut contents = String::new();
+        file.read_to_string(&mut contents)?;
+
+        let config: Config = toml::from_str(&contents)?;
+
+        println!("{:#?}", config);
+
+
+        Ok(config)
+    }
+}
+
+impl Template {
+    use globset::Glob;
+
+    pub fn should_parse(&self, path: Path) -> Result<bool, failure::Error> {
+        match self {
+            Template {
+                include: Some(includes),
+                exclude: None,
+                ..
+            } => check_against_whitelist(path, includes),
+            Template {
+                exclude: Some(excludes),
+                include: None,
+                ..
+            } => check_against_blacklist(path, excludes),
+            _ => unreachable!("Precondition violated: include and exclude are mutually exclusive")
+        }
+    }
+
+    fn check_against_whitelist(path: Path, includes: Vec<String>) -> Result<bool, failure::Error> {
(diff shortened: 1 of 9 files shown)


src/include_exclude.rs
```diff
@@ -0,0 +1,94 @@
...
+                .strip_prefix(project_dir)
+                .expect("strip project dir before matching");
+            !is_dir(e) && !is_git_metadata(e) && should_include(relative_path)
+        } else {
+            false
+        }
+    }))
+}
+
+pub fn create_default_matcher(
```
xortive: I'd like to figure out how to combine this with the create_matcher method, so it returns a single iterator, but the owned closure passed into filter makes this difficult. Would appreciate some help with making it happen!

ashleygwilliams:
would very much like feedback from @k0pernicus and @DD5HT on this!

k0pernicus:
No problem, I can look at the PR this evening :)

src/template.rs
```diff
@@ -119,52 +120,57 @@ pub fn substitute(
...
     let engine = engine();
 
-    for entry in WalkDir::new(project_dir) {
-        let entry = entry?;
-        if is_dir(&entry) || is_git_metadata(&entry) {
-            continue;
+    //returning iterators is hard :/
+    match template_config {
+        Some(template_config) => {
+            for entry in create_matcher(&template_config, project_dir)? {
```
xortive: Clippy is angry about this line and 136 on latest stable. It's a bug in clippy:  https://github.com/rust-lang/rust-clippy/issues/4133

src/config.rs
```diff
@@ -0,0 +1,62 @@
...
+    pub include: Option<Vec<String>>,
+    pub exclude: Option<Vec<String>>,
+}
+
+impl Config {
+    pub fn new<P: AsRef<Path>>(path: P) -> Result<Option<Self>, failure::Error> {
+        match File::open(path) {
+            Ok(mut file) => {
+                let mut contents = String::new();
+                file.read_to_string(&mut contents)?;
```
steveklabnik: https://doc.rust-lang.org/std/fs/fn.read_to_string.html 😄

xortive:
needs squash before merge

src/config.rs
xortive: TIL. thanks!

xortive:
renamed config file name to `cargo-generate.toml` from `.gen.toml`. I am not attached to either name, but I believe the new one is a little better.

src/config.rs
```diff
@@ -0,0 +1,62 @@
+use serde::Deserialize;
+use std::fs::read_to_string;
```
steveklabnik: style nit: for types, like `use serde::Deserialize;`, importing the type fully makes sense. However, for functions, the convention is to use the module, so

    use std::fs;

here, and then below, use `fs::read_to_string`.

src/include_exclude.rs
```diff
@@ -0,0 +1,76 @@
...
+
+pub struct Matcher {
+    template_config: Option<TemplateConfig>,
+    include_matcher: Option<Gitignore>,
+    exclude_matcher: Option<Gitignore>,
+}
+
+impl Matcher {
+    pub fn new(
+        template_config: Option<TemplateConfig>,
```
steveklabnik: Given that this function is one big `if let`/`else`, I would split it into two:

* `new` that only takes `project_dir`
* `new_with_template_config` that takes both a `project_dir` and a `TemplateConfig`

xortive: and leave the option matching to the client instead? makes sense

src/config.rs
xortive: will do!

xortive:
implemented steve's PR feedback, rebased onto master, and added more detail to readme about how this feature interacts with template tags in filenames.

PR #183, opened 2019-08-09
Add warning if config has include and exclude

xortive:
```
🌀  🐑  Generating a new webpack worker project with name 'test504'...
🔧   Creating project called `test504`...
⚠️   Your cargo-generate.toml contains both an include and exclude list. Only the include list will be considered. You should remove the exclude list for clarity. ⚠️
Removed: "/Users/malonso/cf-repos/scratch/test504/.genignore"
Removed: "/Users/malonso/cf-repos/scratch/test504/cargo-generate.toml"
Removed: "/Users/malonso/cf-repos/scratch/test504/LICENSE_APACHE"
Removed: "/Users/malonso/cf-repos/scratch/test504/LICENSE_MIT"
✨   Done! New project created /Users/malonso/cf-repos/scratch/test504
```

warn emoji renders correctly in terminal

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -136,13 +136,13 @@ fn progress(
 ) -> Result<(), failure::Error> {
     let template = template::substitute(name, force)?;
 
-    let pbar = progressbar::new();
-    pbar.tick();
-
     let mut config_path = dir.clone();
     config_path.push(CONFIG_FILE_NAME);
 
     let template_config = Config::new(config_path)?.map(|c| c.template);
+
+    let pbar = progressbar::new();
+    pbar.tick();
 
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -136,14 +136,14 @@ fn progress(
...
 
-    let pbar = progressbar::new();
-    pbar.tick();
-
     let mut config_path = dir.clone();
     config_path.push(CONFIG_FILE_NAME);
 
     let template_config = Config::new(config_path)?.map(|c| c.template);
 
+    let pbar = progressbar::new();
```
xortive: put this after the template config since it swallows stdout, and I don't want to pass pbar to config constructor

PR #194, opened 2019-10-02
Add option to function like `cargo init`

### LINKED ISSUE #193: Add option to function like `cargo init`
Cargo provides two commands for creating a new Rust project: `cargo new` and `cargo init`. The former creates a new subdirectory for the project and a SCM repo. The latter assumes the user has already created the project directory and maybe even the SCM repo and only adds the rest.

`cargo generate` works like `cargo new`, but sometimes the user has already created a Git repo and cloned it so they don't want to have to jump through extra hoops to get everything squared away.

```shell
git clone git@...:myproject.git
cd myproject.git
cargo generate --git git@...:mytemplate.git
ls
# :scream: Sacrébleu, mise en abyme! (and not Inception as most people say but that's an issue for a metaphysical repo)
mv myproject/* .
ls
ls -a
# :dizzy_face: why doesn't * pick up dot files?
mv myproject/.* .
ls -a
git status
# :skull: I overwrote the local repo
```

punkstarman:
closes #193

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -161,6 +161,8 @@ pub struct Args {
     /// Specify the VCS used to initialize the generated template.
     #[structopt(long, default_value = "git")]
     pub vcs: Vcs,
+    #[structopt(long = "bare")]
+    pub bare: bool,
 }
 
 //
@@ -207,6 +209,7 @@ pub fn generate(mut args: Args) -> Result<()> {
 
 fn create_git(args: Args, name: &ProjectName) -> Result<()> {
     let force = args.force;
+    let bare = args.bare;
     let template_values = args
         .template_values_file
         .as_ref()
@@ -220,10 +223,12 @@ fn create_git(args: Args, name: &ProjectName) -> Result<()> {
         args.branch.to_owned(),
     )?;
 
-    if let Some(dir) = &create_project_dir(&name, force) {
+    if let Some(dir) = &create_project_dir(&name, force, bare) {
         match git::create(dir, git_config) {
             Ok(branch) => {
-                git::remove_history(dir)?;
+                if !bare {
+                    git::remove_history(dir)?;
+                }
                 progress(name, &template_values, dir, &branch, &args)?;
             }
             Err(e) => anyhow::bail!(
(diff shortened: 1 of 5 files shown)


punkstarman:
The Windows build isn't working. How can I debug this on a Linux machine?

punkstarman:
Windows!!! :dizzy_face: :fist_raised: :cloud: *raises fist towards the sky in anguish*

punkstarman:
Windows!!! again

src/lib.rs
```diff
@@ -161,6 +161,8 @@ pub struct Args {
     /// Specify the VCS used to initialize the generated template.
     #[structopt(long, default_value = "git")]
     pub vcs: Vcs,
+    #[structopt(long = "bare")]
```
sassman, line 164: I find the name of this argument a bit confusing / misleading.
Git has a `--bare` option that initialises a bare repo, one that can be used as a `remote` directory for push.

Also I'm missing the doc blocks above this flag, that would explain the purpose of this flag to a user (when calling `--help`)

sassman: maybe `--skip-git-init` would express the purpose a bit better?

punkstarman: Good remark about the missing doc block. I just added it.

The purpose is not to skip git init. That has been dealt with by other issues (see https://github.com/cargo-generate/cargo-generate/issues/193#issuecomment-779390079). Rather, the purpose is to be able to avoid creating a directory for the project. It's the same difference between `cargo new` and `cargo init`.

From the beginning, I haven't been sure as to what to call the option (see https://github.com/cargo-generate/cargo-generate/issues/193#issuecomment-537634814 to which there was never a response). The inspiration for `--bare` comes from Haskell Stack (see https://github.com/commercialhaskell/stack/issues/965). Just like you, I wasn't entirely convinced because of the possible confusion with `git init --bare`.

Looking around and thinking about it a bit more, the idea of `--skip-directory` appeals to me. @sassman, what do you think?

What do others think? I am more than open to getting more opinions, as long as we avoid bikeshedding.

sassman: thanks for pointing this out. Now I know why I got confused: `cargo init fo` would actually also create a folder `fo` in `$CWD`. But when the last argument `fo` would be omitted it initializes in `$CWD`. 

So this would actually than mean something like `--into-current-dir` or so. I find the connotation of `skip` then rather misleading. Because it will not skip anything but rather it will create something right in here, in this current working directory.

punkstarman: How about `use-current-dir`?

taurr:
@sassman  I guess this PR can be closed?

taurr:
Ohhh... and are we soon ready for another release? (probably a big one this time)

PR #198, opened 2019-10-23
Fail on failures

EverlastingBugstopper:
This PR changes two instances where cargo-generate would "fail" but exit with code 0.


k0pernicus:
Looks good to me :+1:

PR #205, opened 2019-11-15
Stop attempting template substitutions for invalid files

EverlastingBugstopper:
Fixes #204

PR #207, opened 2019-11-21
🕴️ 0.5.1

EverlastingBugstopper:


k0pernicus:
Blocked by https://github.com/ashleygwilliams/cargo-generate/pull/205 for now.
@EverlastingBugstopper Is there an open discussion to know why those changes imply new alpha version, and not a fix version (for example 0.5.1, as you made before)?

k0pernicus:
Since https://github.com/ashleygwilliams/cargo-generate/pull/237 has been merged into master, I think that we can officialy launch 0.6.0 now.

https://github.com/ashleygwilliams/cargo-generate/pull/205 could be merged and used in another tag version.

EverlastingBugstopper:
sgtm!

EverlastingBugstopper:
i've pushed an update that removes #205 and adds #237 to the changelog - let me know if we think this should be 0.5.1 instead

k0pernicus:
@EverlastingBugstopper I would say that, as #237 introduces new versions of existing libraries, we are not safe from spotting bugs inherent to these libraries, which may introduce weird behavior(s) for cargo-generate...
Tag those changes can be a chance for us to debug in asking to reproduce in older version, and check the difference.

What do you think about this idea?

EverlastingBugstopper:
> Tag those changes can be a chance for us to debug in asking to reproduce in older version, and check the difference.

im not sure what you mean here

k0pernicus:
Just a thing, should I update each MR merged with a milestone tag to 0.6.0 (or 0.5.1) to include them in the changelog?
To be honest, I don't really know how this works... is this automatic or do you have to update it by hand?

k0pernicus:
> > Tag those changes can be a chance for us to debug in asking to reproduce in older version, and check the difference.
> 
> im not sure what you mean here

I misunderstood your question actually... sorry.
I am ok with 0.5.1 :+1:

EverlastingBugstopper:
closing in favor of #246

PR #209, opened 2019-12-26
allows repo name for --git [URL] instead of the entire url

DevinR528:
This allows users to instead of using the entire url to only type the name of the repo/project. I realize this is only partially useful as most people will simply paste the url from the TEMPLATES.md file. That being said if this seems useful cool, otherwise I understand completely.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -15,6 +15,7 @@ use cargo;
 use config::{Config, CONFIG_FILE_NAME};
 use console::style;
 use failure;
+use std::collections::HashMap;
 use std::env;
 use std::path::PathBuf;
 use structopt::StructOpt;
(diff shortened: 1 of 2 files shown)


k0pernicus:
Hi @DevinR528,
Thank you for your contribution :)  
I am not super confident about this feature, as we will have to manage by hand all the projects, based on the solution you are proposing, but also based on possible hashmap conflicts that can happen later...  
Also, as you said, most people simply paste the url from the TEMPLATES.md file.

@ashleygwilliams Are you interested by the feature right now?

sassman:
@DevinR528 @k0pernicus this could be a bit refactored and combined with #210 in a way that a default config file with urls provided in this PR (from TEMPLATES.md) can be shipped. 

So that every user can modify this on his local config folder and add their own templates.

What do you think? Of course that might require changes on the design, leaving the hash map behind and reading a config file instead.

DevinR528:
I think that sounds like a good idea. I unfortuantly don't have the time now to work on it :( but may in a few months...
Thanks for letting me know about the other issue.

SirWindfield:
I simple solution would be to check if the url is of pattern <username>/<repo> and create a url for github using format!. I can work on this later on if this is still desired.

k0pernicus:
@SirWindfield this requires that everyone is using github to host their projects... and not a personal bitbucket or gitlab instance for example.  
Maybe it's the case for today, but maybe not in the future.

The solution proposed by @sassman sounds better for me imoo.

SirWindfield:
Faie enough, I guess this can be closed then

k0pernicus:
Can we re-open your issue @SirWindfield (https://github.com/ashleygwilliams/cargo-generate/issues/231) and discuss about a solution there?

I do think we can close this PR for the moment, indeed.

SirWindfield:
There was #210 which was basically talking about the same as mine (#231). That's why I closed mine :)

k0pernicus:
Ok good, thanks @SirWindfield

k0pernicus:
Hi @DevinR528 

Do you still want to work on this issue?
I can move it to "draft" (instead of "open") if you want to work on later :)

DevinR528:
Yeah, I'm fine with closing this PR. If I have more time #210 seems reasonable to implement but don't wait for if anyone else wants to tackle it.

k0pernicus:
Thanks again for your work @DevinR528 💪 👍

PR #218, opened 2020-03-20
Add template-rust-backend-with-electron-frontend

usagi:
The new template is here: https://github.com/usagi/template-rust-backend-with-electron-frontend


k0pernicus:
Thanks for your contribution :)

PR #220, opened 2020-04-13
Add Win32 template

ArmsOfSorrow:
I've been playing with the Windows API lately and decided to create a template that displays a blank window and runs a message loop as a starting point.

You can include it in the list if you want. :)


k0pernicus:
Awesome! Thank you @ArmsOfSorrow :)

PR #242, opened 2020-08-26
Support abbreviated `owner/repo` Git URL format

9999years:
The GitHub [hub] tool allows using `owner/repo` as an abbreviated form of `https://github.com/owner/repo.git`. This change supports that format in the `--git` command-line argument.

It also improves the command-line argument documentation.

[hub]: https://github.com/github/hub


src/git.rs
```diff
@@ -16,29 +16,47 @@ pub struct GitConfig {
...
             remote,
             branch: GitReference::Branch(branch),
         })
     }
+
+    /// Creates a new `GitConfig`, first with `new` and then as a GitHub `owner/repo` remote, like
+    /// [hub].
+    ///
+    /// [hub]: https://github.com/github/hub
+    pub fn new_abbr(git: &str, branch: String) -> Result<Self, failure::Error> {
```
k0pernicus, line 55: I'm a little embarrassed with this as we definitely become more and more closer to github and not providers like gitlab, atlassian products, etc...

k0pernicus (commented):
Will need another review.

@DD5HT, @ashleygwilliams ?

src/git.rs
9999years: Yeah, it's... unfortunate. Both because of GitHub's awful politics (drop ICE, you bastards) and because we've effectively centralized a purposefully distributed VCS. (Not like Atlassian's [any good](https://twitter.com/mcclure111/status/12770025523039027270) either, though.)

But also... it's a very convenient feature that helps improve usability. It might be nice to have a (hardcoded...?) name-to-URL map as well, for the projects in TEMPLATES.md.

9999years: And, just pragmatically, every single template currently listed in TEMPLATES.md is on GitHub anyways.

k0pernicus: > And, just pragmatically, every single template currently listed in TEMPLATES.md is on GitHub anyways.

Yep but, in my own opinion, this does not have to push us to "close" some doors...

Let's wait if other maintainers have an opinion on that or not (if they come).

sassman: Just a thought on this, right now the `--git` cli param defaults to github, which is fine. What about introducing others like `--gitlab` or `--bitbucket` that behave as `--git` just with a different url string that is used here in line 57

k0pernicus: Thanks @sassman - this can indeed solve some issues...

We can register some "providers" (github, gitlab, bitbucket, ...), switch on the one provided by the user and use it next.

9999years: I would support that, but remember that the good is the enemy of the perfect and this code is here, working and tested, *now.* Of course I'll defer to the maintainers but I don't know if this is worth bikeshedding that much over.

sassman: IMO it's an improvement anyway, means the old style with a full repo URL is still available. 

So I would vote for merge as it is and wait for feedback / wishes from users after releasing it.

sassman: Hey there, can we just land this? I don't see any reason to wait any longer. 
Also we should plan a new release.. I guess it's time.

k0pernicus: As it seems the original maintainers don't care about the project, I think we can bypass their approval indeed.

sassman: Actually I would not call it baypassing, if something goes in a direction they don't want, then they surely would jump in and make decissions, not seeing them making any statements for means just they agree with everything. 😁

k0pernicus:
I resolved the current conflicts with master.
I am now waiting for the tests to run, before merging to master.

k0pernicus:
Done, thanks again for your work :)

PR #245, opened 2020-09-13
Feat/graceful handle unknown variables and other errors during parsing

sassman:
@k0pernicus this is a take to solve #205 

It's a bit hacky since liquid has no great way to figure out what's the exact problem or to ignore undefined variables. But eventually this should do the trick. If you see more test cases that we should cover feel free to point them out.

### DIFF
--- a/src/template.rs
+++ b/src/template.rs
@@ -203,7 +203,8 @@ fn construct_substition_warning(files_with_errors: Vec<(String, liquid::error::E
             msg.push_str("\n");
         }
         msg.push_str("\n");
-        msg.push_str("Consider adding these files to a `cargo-generate.toml` to skip substition on these files.\n");
+        let info = format!("{}", style("Consider adding these files to a `cargo-generate.toml` to skip substition on these files.\n").bold());
+        msg.push_str(&info);
         msg.push_str("Learn more: https://github.com/ashleygwilliams/cargo-generate#include--exclude.\n\n");
         msg
 }
--- a/src/template.rs
+++ b/src/template.rs
@@ -9,10 +9,11 @@ use heck::{CamelCase, KebabCase, SnakeCase};
 use indicatif::ProgressBar;
 use liquid_core::{Filter, FilterReflection, Object, ParseFilter, Runtime, ValueView, Value};
 use std::env;
+use liquid::Template;
+use liquid_core::{Filter, FilterReflection, Object, ParseFilter, Runtime, Value, ValueView};
 use std::fs;
 use std::path::Path;
 use walkdir::{DirEntry, WalkDir};
-use liquid::Template;
 
 fn engine() -> liquid::Parser {
     liquid::ParserBuilder::with_stdlib()
@@ -174,7 +175,11 @@ pub(crate) fn walk_dir(
     Ok(())
 }
 
-fn parse_and_replace(context: &Object, filename: &Path, parsed_file: Template) -> Result<(), anyhow::Error> {
+fn parse_and_replace(
+    context: &Object,
+    filename: &Path,
+    parsed_file: Template,
+) -> Result<(), anyhow::Error> {
     let new_content = parse_file(context, filename, &parsed_file);
     fs::write(filename, new_content).with_context(|| {
         format!(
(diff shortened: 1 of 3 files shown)


src/template.rs
```diff
@@ -148,29 +151,96 @@ pub(crate) fn walk_dir(
...
+                Ok(parsed_file) => parse_and_replace(&template, filename, parsed_file)?,
+                Err(e) => {
+                    if verbose {
+                        files_with_errors.push((filename.display().to_string(), e.clone()));
+                    }
+                }
+            }
         }
     }
+    if files_with_errors.len() > 0 {
```
Veetaha, line 165:
```suggestion
    if !files_with_errors.is_empty() {
```

src/template.rs
```diff
@@ -148,29 +151,96 @@ pub(crate) fn walk_dir(
...
 
     pbar.finish_and_clear();
     Ok(())
 }
+
+fn parse_and_replace(
+    context: &Object,
+    filename: &Path,
+    parsed_file: Template,
+) -> Result<(), anyhow::Error> {
```
Veetaha, line 178:
```suggestion
) -> anyhow::Result<()> {
```

src/template.rs
```diff
@@ -148,29 +151,96 @@ pub(crate) fn walk_dir(
...
+            .red(),
+    );
+    for file_error in files_with_errors {
+        msg.push_str("\t");
+        msg.push_str(&file_error.0);
+        msg.push_str("\n");
+    }
+    msg.push_str("\n");
+    let info = format!("{}", style("Consider adding these files to a `cargo-generate.toml` in the template repo to skip substitution on these files.\n").bold());
+    msg.push_str(&info);
```
Veetaha, line 241: This requires `use std::fmt::Write` at the top of the file, but:
```suggestion
    writeln!(msg, "{}", style("Consider adding these files to a `cargo-generate.toml` in the template repo to skip substitution on these files.").bold()).unwrap();
```

src/template.rs
```diff
@@ -148,29 +151,96 @@ pub(crate) fn walk_dir(
...
+                // so, we miss a variable that is present in the file to render
+                let requested_var =
+                    regex::Regex::new(r"(?P<p>.*requested\svariable=)(?P<v>.*)").unwrap();
+                let captures = requested_var.captures(msg.as_str()).unwrap();
+
+                if let Some(Some(req_var)) = captures.iter().last() {
+                    let missing_variable = req_var.as_str().to_string();
+                    // try again with this variable added to the context
+                    let mut context = context.clone();
+                    context.insert(missing_variable.into(), Value::scalar("".to_string()));
```
sassman, line 207: this is something to be discussed, one could argue to keep the variable as it is provided. So better might be to substitute it with something like `format!("{{ {} }}", missing_variable.into())`

src/template.rs
sassman: unfortunately this won't compile:

```
error[E0599]: no method named `write_fmt` found for struct `std::string::String` in the current scope
   --> src/template.rs:236:5
    |
236 |     writeln!(msg, "{}", style("Consider adding these files to a `cargo-generate.toml` in the template repo to skip substitution on these files.").bold()).un...
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ method not found in `std::string::String`
    |
    = help: items from traits can only be used if the trait is in scope
    = note: the following trait is implemented but not in scope; perhaps add a `use` for it:
            `use std::fmt::Write;`
    = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `cargo-generate`.
```

k0pernicus: @sassman I think you forgot the `use std::fmt::Write` as Veetaha stated before :)

sassman: sorry my bad. fixed.

src/template.rs
sassman: It might be even possible to promt the user an interactive input for defining this variable. So that the decision is delegated to the user and if he not wants this variable filled he can leave it empty and move on.

@k0pernicus EverlastingBugstopper any opinions on that?

sassman:
coming back to this PR after a while I would say that the safest way to handle unknown placeholders is to keep them as they are. Presenting a warning message, but then keeping them as they appeared would be the less harmful way.

So I will change the replacement of unknown placeholders the way that they will be replaces by itself as they are, since that is the only way to convince liquid to not break.

PR #263, opened 2020-12-06
Handle default branch properly

cecton:
Hi,

Thank you for `cargo-generate` :heart: 

I would like to introduce this change to make `cargo-generate` more friendly with repositories that don't use "master" as the default branch. I think this is particularly relevant now that GitHub also uses "main" instead of "master".

I would have like to fix it on cargo directly but there are some compatibility issues. But I don't think this is an issue for `cargo-generate`. This is why I'm introducing this change here.

I was personally bothered by it because I tried to make a new template but it failed because I was using the "main" branch as default branch.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -131,9 +131,9 @@ checksum = "81a18687293a1546b67c246452202bbbf143d239cb43494cc163da14979082da"
 
 [[package]]
 name = "cargo"
-version = "0.46.1"
+version = "0.49.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


k0pernicus (approved):

k0pernicus:
Hi @cecton, thanks for the update!

Everything is ok on my side - any comment(s) @sassman or I can merge it? :)

sassman:
@k0pernicus unfortunately I cannot merge it, lacking the privileges. But looks good for me.

k0pernicus:
Thanks again @cecton, and welcome as a contributor! :)

cecton:
Thanks a lot! 🎉

PR #264, opened 2020-12-17
make Args fields public to enable generate to be called directly from external code

No9:
Hey All 
First of all thanks for a very useful tool.

I am using this project as a library and would like to be able to create an `Args` struct and pass it to the `generate` function without having to create a cli wrapper. 
I have found a work-around using `std::mem::transmute` so I don't have to carry a fork of this project but thought you might be interested in a PR that enables this capability.
The change is fairly trivial and I have provided a copy of one of the more compete tests to validate but didn't want to duplicate all the test stack as it felt like  too much over head for little added benefit on top of what is already covered in `basic.rs`

Hope that all makes sense.


macalimlim:
this would be great because instead of depending on an external tool, that tool is integrated to your own tool as a library. A good use case for this is with a tool I created https://crates.io/crates/ftw

k0pernicus:
Thanks @No9, I don't see any reason to not accept the PR...

Do you anything wrong here @sassman ? :)

sassman (approved):

sassman:
@k0pernicus looks all good.

k0pernicus (approved):

sassman:
Thank you @No9 for your contribution 💪

No9:
Thanks @sassman and @k0pernicus for landing this.

PR #267, opened 2020-12-27
docs: fix command name

MarcoIeni:
`cargo gen` is a typo, right?


MarcoIeni:
I just saw that it exists, but on my system is not recognized.

```
$ cargo gen                    
error: no such subcommand: `gen`

	Did you mean `new`?
```

PR #269, opened 2021-01-01
Ignore .cargo-ok files by default.

### LINKED ISSUE #259: Creates an empty `.cargo-ok` file
```console
> cargo generate --git https://github.com/Michael-F-Bryan/github-template --name const-arrayvec
> ls const-arrayvec/.cargo-ok
const-arrayvec/.cargo-ok
> cargo generate --version
cargo-generate 0.5.1
```

This file causes issues for `cargo` (though that should be fixed in 1.49) https://github.com/rust-lang/cargo/issues/8816

thomcc:
I was going to write a script that wraps `cargo-generate` and then automatically deletes this, but it's pretty easy to just add to the default ignore set.

This file serves no purpose and causes clutter, and many users either fail to notice it or think it should be kept: https://sourcegraph.com/search?q=file:.cargo-ok&patternType=literal

I suspect any case where this file is actually part of the template is either broken or at best fragile (as it's existence can prevent cargo from detecting failure to extract the tarball), depending on cargo's behavior. (That said, if any do exist, it seems very likely they came into existence by something like using `cargo-generate` to generate a template repo).

Fixes #259

### DIFF
--- a/src/ignoreme.rs
+++ b/src/ignoreme.rs
@@ -3,12 +3,21 @@ use remove_dir_all::*;
 use std::fs::remove_file;
 use std::{
     collections::HashSet,
-    iter,
     path::{Path, PathBuf},
 };
 
 use crate::config::CONFIG_FILE_NAME;
 pub(crate) const IGNORE_FILE_NAME: &str = ".genignore";
+
+// We ignore the `.cargo-ok` file if one is present. This file is a somewhat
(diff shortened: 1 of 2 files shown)


k0pernicus (approved):

k0pernicus:
Thanks @thomcc, the changes look good to me!

PR #275, opened 2021-01-20
Enable usage of custom keywords

pedrohjordao:
Work towards closing #17

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,3 +1,83 @@
+//! Generate a new Cargo project from a given template
+//!
+//! Right now, only git repositories can be used as templates. Just execute
+//!
+//! $ cargo generate --git https://github.com/user/template.git --name foo
+//!
+//! or
+//!
+//! $ cargo gen --git https://github.com/user/template.git --name foo
+//!
+//! and a new Cargo project called foo will be generated.
+//!
+//! TEMPLATES:
+//!
+//! In templates, the following placeholders can be used:
+//!
+//! - `project-name`: Name of the project, in dash-case
+//!
+//! - `crate_name`: Name of the project, but in a case valid for a Rust
+//!   identifier, i.e., snake_case
+//!
+//! - `authors`: Author names, taken from usual environment variables (i.e.
+//!   those which are also used by Cargo and git)
+//!
+//! The template author can define their own placeholders in their
+//! `cargo-generate.toml` file. This looks like the following:
+//!
+//! ```toml
+//!
+//! [placeholders]
+//!
+//! my-placeholder = { type = "string", prompt = "Hello?", choices = ["hello", "world"], default = "hello", regex = "*" }
+//!
+//! use-serde = { type = "bool", prompt = "Add serde support?", default = false }
+//!
+//! ```
+//!
+//! The user of the template will then be asked the the question in "prompt", and must accept the
+//! default value (if provided) or enter a custom value which will be checked against "choices" (if
+//! provided) and regex (if provided).
+//!
+//! The placeholder map supports the following keys:
+//!
+//! `type` (required): Must be "string" or "bool"
+//!
+//! `prompt` (required): A string containing the question to be asked to the user
+//!
+//! `default` (optional): The default value to be used if the user just presses enter. Must be
+//! consistent with `type`
+//!
+//! `choices` (optional; string only): Possible values the user may enter
+//!
+//! `regex` (optional; string only): Regex to validate the entered string
+//!
+//! For automation purposes the user of the template may provide provide a file containing the
+//! values for the keys in the template by using the `--template-values-file` flag.
+//!
+//! The file should be a toml file containing the following (for the example template provided above):
+//!
+//! ```toml
+//!
+//! [values]
+//!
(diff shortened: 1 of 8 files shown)


sassman:
Hey @pedrohjordao,

this looks really good so far. Thanks for your hard work. I think within the next days I'm able to finish the review and testing.
I would also like to overhaul / enrich the documentation, so maybe a feature branch would help to keep others contribution to it.

src/lib.rs
```diff
@@ -41,6 +45,48 @@ use structopt::StructOpt;
 ///
 /// - `authors`: Author names, taken from usual environment variables (i.e.
 ///   those which are also used by Cargo and git)
+///
+/// The template author can define their own placeholders in their
+/// `cargo-generate.toml` file. This looks like the following:
+///
+/// [placeholders]
```
sassman, line 52: this embedded code needs to be wrapped inside a `code` block that starts with `` ```toml ``

you can verify that it works via `cargo doc --verbose` then you see the warnings.

src/lib.rs
```diff
@@ -41,6 +45,48 @@ use structopt::StructOpt;
...
+/// my-placeholder = { type = "string", prompt = "Hello?", choices = ["hello", "world"], default = "hello", regex = "*" }
+/// use-serde = { type = "bool", prompt = "Add serde support?", default = false }
+///
+/// The user of the template will then be asked the the question in "prompt", and must accept the
+/// default value (if provided) or enter a custom value which will be checked against "choices" (if
+/// provided) and regex (if provided).
+///
+/// The placeholder map supports the following keys:
+///
+/// `type` [required]: Must be "string" or "bool"
```
sassman, line 62: those brackets in `[required]` hints cargo doc to create a link to a field of a struct, but here we don't reference a field from a struct so it's required to either mask the brackets or remove / replace them with something else.

src/lib.rs
pedrohjordao: Yeah, right

The issue is how it looks like in the generated help message, but I'll come up with something

src/lib.rs
pedrohjordao: thanks, I'll fix it

src/lib.rs
sassman: I suggest we move it all to a module level comment `//!` then it appears on the docs quite nicely.

pedrohjordao: Right, but then structop isn't able to pick it up as the help message, right? Am I missing something?

sassman: I'm not super familiar with `structops` but I was under the impression that the help messages comes directly from the comments above the individual fields. But I can double-check it.

src/projectvariables.rs
```diff
@@ -0,0 +1,835 @@
...
+    String(String),
+}
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+enum SupportedVarType {
+    Bool,
+    String,
+}
+
+const INVALID_NAMES: [&str; 2] = ["project-name", "crate_name"];
```
sassman, line 80: Maybe the other standard variables should be named here too? Like: `os-arch` and `authors`

src/projectvariables.rs
```diff
@@ -0,0 +1,835 @@
+use liquid_core::Value;
```
sassman, line 1: I would suggest renaming this file to have a `_` in the name, like `project_variables.rs`

sassman (commented):
Found only few nits, over all very good work! Thanks again.

src/lib.rs
pedrohjordao: AFAIK, it gets the help message from the struct documentation itself for the general `-h` message

You can see it if you just do `cargo run` with no extra parameters

src/projectvariables.rs
pedrohjordao: makes sense to me

sassman:
Could you also please rebase?

pedrohjordao:
Should I squash the changes also?

sassman:
That would be great, if possible into one commit with a nice commit message that sums it up.

src/lib.rs
```diff
@@ -113,6 +196,20 @@ fn create_git(args: Args, name: &ProjectName) -> Result<()> {
...
 
+fn get_config_file_values(path: &Path) -> Result<HashMap<String, toml::Value>> {
+    match fs::read_to_string(path) {
+        Ok(ref contents) => toml::from_str::<ConfigValues>(contents)
+            .map(|v| v.values)
+            .map_err(|e| e.into()),
+        Err(e) => anyhow::bail!(
+            "{} {} {}",
+            emoji::ERROR,
+            style("VAlues File Error:").bold().red(),
```
sassman, line 207: here is a little typo, `VAlues` but should be `Value File Error` possibly

sassman (changes requested):
sorry found one more little thing. But then it looks good.

src/lib.rs
pedrohjordao: Done

sassman (approved):
Nice job

sassman:
could you please do a rebase on master?

PR #290, opened 2021-02-08
dont init git when inside existing repo

taurr:
When generating templates inside an existing repository, there is no need to attempt the initialization of a new one - in fact it just makes it harder to integrate the new template!

### DIFF
--- a/src/git.rs
+++ b/src/git.rs
@@ -117,10 +117,12 @@ pub(crate) fn remove_history(project_dir: &Path) -> Result<()> {
 }
 
 pub fn init(project_dir: &Path, branch: &str) -> Result<GitRepository> {
-    let mut opts = RepositoryInitOptions::new();
-    opts.bare(false);
-    opts.initial_head(branch);
-    GitRepository::init_opts(project_dir, &opts).context("Couldn't init new repository")
+    GitRepository::discover(project_dir).or_else(|_| {
+        let mut opts = RepositoryInitOptions::new();
+        opts.bare(false);
+        opts.initial_head(branch);
+        GitRepository::init_opts(project_dir, &opts).context("Couldn't init new repository")
+    })
 }
 
 #[cfg(test)]
--- a/tests/integration/basics.rs
+++ b/tests/integration/basics.rs
@@ -1,6 +1,6 @@
 use predicates::prelude::*;
 
-use crate::helpers::project_builder::dir;
+use crate::helpers::project_builder::tmp_dir;
 
 use assert_cmd::prelude::*;
 use std::env;
@@ -13,7 +13,7 @@ fn binary() -> Command {
 
 #[test]
 fn it_substitutes_projectname_in_cargo_toml() {
-    let template = dir("template")
+    let template = tmp_dir()
         .file(
             "Cargo.toml",
             r#"[package]
@@ -25,7 +25,7 @@ version = "0.1.0"
         .init_git()
         .build();
 
-    let dir = dir("main").build();
+    let dir = tmp_dir().build();
 
     binary()
         .arg("generate")
@@ -47,7 +47,7 @@ version = "0.1.0"
 
 #[test]
 fn it_substitutes_date() {
-    let template = dir("template")
+    let template = tmp_dir()
         .file(
             "Cargo.toml",
             r#"[package]
@@ -59,7 +59,7 @@ version = "0.1.0"
         .init_git()
         .build();
 
-    let dir = dir("main").build();
+    let dir = tmp_dir().build();
 
     binary()
         .arg("generate")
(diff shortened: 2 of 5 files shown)


src/git.rs
```diff
@@ -117,10 +117,17 @@ pub(crate) fn remove_history(project_dir: &Path) -> Result<()> {
 }
 
 pub fn init(project_dir: &Path, branch: &str) -> Result<GitRepository> {
-    let mut opts = RepositoryInitOptions::new();
-    opts.bare(false);
-    opts.initial_head(branch);
-    GitRepository::init_opts(project_dir, &opts).context("Couldn't init new repository")
+    if let Some(repo) = project_dir
```
sassman, line 120: Nice idea! Could you also please add a little test case for this case?

taurr:
@sassman 
I've added a test for generating a template inside an already generated template - ensuring that the inner one isn't initialized as a git repository.

taurr:
Just discovered a better way of  doing this...

tests/integration/helpers/project_builder.rs
```diff
@@ -7,23 +7,40 @@ use std::sync::atomic::*;
...
 pub struct ProjectBuilder {
     files: Vec<(String, String)>,
     submodules: Vec<(String, String)>,
-    root: PathBuf,
+    root: RootType,
     git: bool,
     branch: Option<String>,
 }
 
+enum RootType {
```
sassman, line 23: I like the idea so much, that I would vote for not introducing this enum, and change `root: RootType` to `root: TempDir` so that we can get rid of the self made tempdir thing that was stored in root before. 

I've checked it locally and then also `impl Drop for Project` should be removed.

sassman: Then `fn root()` would shrink to something like 

```rust
fn root(name: &str) -> TempDir {
    tempdir().unwrap()
}
```

tests/integration/helpers/project.rs
```diff
@@ -3,9 +3,12 @@ use std::io::Read;
 use std::path::{Path, PathBuf};
 use std::str;
 
+use tempfile::TempDir;
+
 #[derive(Debug)]
 pub struct Project {
     pub root: PathBuf,
+    pub(crate) _tmpdir: Option<TempDir>,
```
sassman, line 11: pls name it just `tmpdir` without the `_`, since it needs to be kept until the test is done, so that the drop impl of `TempDir` would cleanup the directory.

tests/integration/helpers/project_builder.rs
sassman: this would also make some of the changes below obsolete.

sassman (changes requested):
Hey, thanks again for the PR, I found a few smaller things that are little improvements on the code base, that would reduce the final overall complexity (mostly on the tests)

I would appreciate if you could have a look at them.

sassman:
At the very end please rebase + squash the commits into one commit with a nice message that sums it all up.

src/git.rs
```diff
@@ -117,10 +117,14 @@ pub(crate) fn remove_history(project_dir: &Path) -> Result<()> {
 }
 
 pub fn init(project_dir: &Path, branch: &str) -> Result<GitRepository> {
-    let mut opts = RepositoryInitOptions::new();
-    opts.bare(false);
-    opts.initial_head(branch);
-    GitRepository::init_opts(project_dir, &opts).context("Couldn't init new repository")
+    if let Ok(repo) = GitRepository::discover(project_dir) {
```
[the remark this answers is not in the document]
taurr: Agreed :-)

tests/integration/basics.rs
```diff
@@ -642,7 +687,7 @@ version = "0.1.0"
 }
 
 #[test]
-fn it_respects_template_branch_name() {
+fn it_respects_template_branch_name() -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
taurr: Leftover from a brief period in time where the test did use the `?` operator.
I Removed it, in the cleanup

tests/integration/basics.rs
```diff
@@ -677,6 +722,7 @@ fn it_respects_template_branch_name() {
         .assert()
         .success()
         .stdout(predicates::str::contains("On branch gh-pages").from_utf8());
+    Ok(())
```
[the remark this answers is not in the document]
taurr: Removed in cleanup

tests/integration/helpers/project_builder.rs
taurr: I also like the idea of just using tempdir, so I went along and cleaned it up - no more 
```rust
fn dir(name: &str) -> ProjectBuilder
```
either... hope its ok.

tests/integration/helpers/project.rs
taurr: obsolete due to other changes!

taurr:
Changes implemented, commits squashed 😃

sassman (approved):
Looks good, thanks a lot.

Issue #210, opened 2020-01-24
Let cargo-generate read a config file with defaults.


najamelan:
@ashleygwilliams I wonder how you feel about the possibility to set defaults, eg. if one often uses the same template, it would be nice to be able to omit the `--git` command line parameter. Also I would like to be able to turn off renaming a project with underscore to hyphen, saving another manual `mv` command. I imagine that other configuration options will come up once the infra is available.

Would you accept a PR for this? I would propose the user dir, something like `~/.config/cargo-generate.toml` and an equivalent on Windows.

ps: Let's be complete in saying that it can be worked around by creating a shell function to wrap cargo-generate, so it's not the end of the world.

EverlastingBugstopper:
I'm not sure about the default git template functionality - that seems like something more suited to a shell alias than an explicit pattern.

As for turning off the renaming, I believe you should be able to do that with the `--force` flag already. This is not documented though - if you try it and it works, it may be worth adding a documentation PR.

sassman:
@najamelan I was thinking about exactly the same. To give a bit more context, I imagined that the config file might contain something like:

```toml
[shortcuts]
cli = "https://github.com/sassman/cli-template-rs.git"
lib = "https://github.com/sassman/lib-template-rs.git"
```

and then I can use it this way:

```sh
cargo generate -s cli -name my-new-cli
# or the long version 
cargo generate --shortcut cli -name my-new-cli
```

SirWindfield:
@sassman I really like that idea! Is there interest in getting something like this merged into `cargo-generate`?

bilelmoussaoui:
I have a similar use case here https://gitlab.gnome.org/bilelmoussaoui/gtk-rust-template/-/issues/7 where I would like to provide a template for gtk3 & gtk4 without having to duplicate a bunch of files so I guess something like this would work somehow?

taurr:
I might have the start for this - see #256

sassman:
Hi @taurr,

your [feature/favorites](https://github.com/cargo-generate/cargo-generate/compare/master...taurr:feature-favorites) is even better than the proposal above. 

Do you want to add a few more tests and contribute with a PR? I would certainly appreciate it.

taurr:
Hi @sassman 

I will add a few more tests tests, then make the PR - that was my intention from the get go, just didn't know how my proposal would be "judged" (first time Github contributer!) 😄

sassman:
[...]

PR #292, opened 2021-02-09
Feature favorites

taurr:
Implements favorites (shortcut) and reading of a configuration file, as partly described in #210

### DIFF
--- a/src/app_config.rs
+++ b/src/app_config.rs
@@ -1,0 +1,57 @@
+use anyhow::{Context, Result};
+use serde::Deserialize;
+use std::{
+    collections::HashMap,
+    fs,
+    path::{Path, PathBuf},
+};
+
+pub(crate) const CONFIG_FILE_NAME: &str = "cargo-generate";
+
+#[derive(Deserialize)]
+pub(crate) struct AppConfig {
+    pub favorites: HashMap<String, FavoriteConfig>,
+}
+
+#[derive(Deserialize)]
+pub(crate) struct FavoriteConfig {
+    pub description: Option<String>,
+    pub git: Option<String>,
+    pub branch: Option<String>,
+}
+
+impl Default for AppConfig {
+    fn default() -> Self {
+        Self {
+            favorites: HashMap::new(),
+        }
+    }
+}
+
+impl AppConfig {
+    pub(crate) fn from_path(path: &Path) -> Result<AppConfig> {
+        if path.exists() {
+            let cfg = fs::read_to_string(path)?;
+            Ok(if cfg.trim().is_empty() {
+                AppConfig::default()
+            } else {
+                toml::from_str(&cfg)?
+            })
+        } else {
+            Ok(Default::default())
+        }
+    }
+}
+
+pub(crate) fn app_config_path(path: &Option<PathBuf>) -> Result<PathBuf> {
+    path.clone()
+        .or_else(|| home::cargo_home().map_or(None, |h| Some(h.join(CONFIG_FILE_NAME))))
+        .with_context(|| {
+            format!(
+                r#"
+Unable to resolve path for configuration file.
+Use --config option, or place {} in $CARGO_HOME."#,
+                CONFIG_FILE_NAME
+            )
+        })
+}
--- a/src/favorites.rs
+++ b/src/favorites.rs
@@ -1,0 +1,68 @@
+use crate::{
+    app_config::{app_config_path, AppConfig, FavoriteConfig},
+    Args,
+};
+use anyhow::{anyhow, Context, Result};
+
+pub(crate) fn list_favorites(args: &Args) -> Result<()> {
+    let path = &app_config_path(&args.config)?;
+    println!("Listing favorites defined in: {}", path.as_path().display());
+    let app_config = AppConfig::from_path(path)?;
+
+    let data = {
+        let mut d = app_config
+            .favorites
+            .iter()
+            .filter(|(key, _)| args.favorite.as_ref().map_or(true, |f| key.starts_with(f)))
+            .collect::<Vec<(&String, &FavoriteConfig)>>();
+        d.sort_by_key(|(key, _)| key.to_string());
+        d
+    };
+
+    println!("Possible favorites:");
+    let longest_key = data.iter().map(|(k, _)| k.len()).max().unwrap_or(0);
+    let longest_key = ((longest_key + 5) / 4) * 4;
+    data.iter().for_each(|(key, conf)| {
+        println!(
+            "{}:{}{}",
+            key,
+            " ".repeat(longest_key - key.len()),
+            conf.description.as_ref().cloned().unwrap_or_default()
+        );
+    });
+
+    Ok(())
+}
+
+pub(crate) fn resolve_favorite(args: &mut Args) -> Result<()> {
+    if args.git.is_some() {
+        return Ok(());
+    }
+
+    let favorite_name = args
+        .favorite
+        .as_ref()
+        .ok_or_else(|| anyhow!("Please specify either --git option, or a predefined favorite"))?;
+
+    let app_config_path = app_config_path(&args.config)?;
+    let app_config = AppConfig::from_path(app_config_path.as_path())?;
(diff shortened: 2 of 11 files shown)


taurr:
@sassman 
Tests has now been added, and the resolving bugs has been fixed (had to spin up a VM in order to do that, would be nice to have one for mac too).
Hope its useable 😄

sassman:
Awesome, I'm on Mac and will test it all soonish. 
Very glad you finished your fist great contribution.

I'll get back to you this week.

src/app_config.rs
```diff
@@ -0,0 +1,52 @@
...
+pub(crate) fn edit_config(path: Option<PathBuf>) -> anyhow::Result<()> {
+    if let Some(path) = path {
+        println!(
+            r#"Attempting to edit config at "{}""#,
+            path.as_path().display().to_string()
+        );
+        if !path.exists() {
+            std::fs::File::create(&path)?;
+        }
+        open::that(path)?;
```
sassman, line 27: I'd prefer to only present the user a message here to along the lines of

`println!("to edit your config: $EDITOR {}", path);`

That way we not need to pull in the `open` crate as a dependency, and keep it lean. Since the target user group is a shell user, we should keep things simple, that that would be directing the user `what to do` and `where to do it`

src/app_config.rs
```diff
@@ -0,0 +1,52 @@
...
+    } else {
+        anyhow::bail!("No configuration file specified")
+    }
+}
+
+pub(crate) fn read_config(path: Option<PathBuf>) -> anyhow::Result<Option<AppConfig>> {
+    match path {
+        Some(path) if path.exists() => {
+            let config_content = fs::read_to_string(&path)?;
+            let config_content = config_content.trim();
```
sassman, line 38: those calls are chainable like `fs::read_to_string(&path)?.trim();`

src/lib.rs
```diff
@@ -68,25 +84,58 @@ pub struct Args {
...
+    }
+
+    let app_config = read_config(app_config_path)?;
+    if args.list_favorites {
+        return list_favorites(&app_config, &args);
+    }
+
+    if let (Some(favorite), None) = (&args.favorite, &args.git) {
+        if let Some(app_config) = app_config {
+            if let Some(favorite) = app_config.favorites.get(favorite.as_str()) {
```
sassman, line 111: could you please rewrite this to avoid the deep nesting of `if let` also I think what would be good refactored into an own function, since this is not exactly related to `generate` more of a look up for the right git url and branch.

sassman (changes requested):
Hey, great stuff. I found only few little nits and I would be very happy if we can avoid the `open` crate as a dependency to keep things small and cli focused (imagine users without X or any GUI)

src/app_config.rs
taurr: The reason I used the `open` crate was actually to get it working on both windows and linux (and I've tested it)

I have no problem getting rid of it again - but that also means the `--edit-config` should be removed.
Then the question is: when/under what option, do we show where the config file is located??

src/app_config.rs
taurr: Nope - doing so makes the compiler complain, as `trim()` only returns a reference and thus we have issues with a temporary variable 😢 
Getting rid of the `trim()` makes the toml parser (serde) complain on an empty config file 😭

Options?

taurr: Fixed - but not where suggested as that would introduce a temporary variable problem

src/app_config.rs
sassman: What about the [`edit crate`](https://docs.rs/edit/0.1.2/edit/)? 
Would also allow to fetch the input the user made directly inline here as return.

taurr: I solved it by entirely removing the --edit-config option, then adding a better error description if/when users try to use the favorite feature with either no config or missing favorites.

So, better error descriptions and getting rid of the dependency.

If at some time we'd like the --edit-config feature, we should probably create a new feature request :-)

taurr:
@sassman 
I've rebased, squashed and tried to solve all your issues above.

src/app_config.rs
```diff
@@ -0,0 +1,59 @@
...
+                }
+            })?)
+        } else {
+            Ok(Default::default())
+        }
+    }
+}
+
+pub(crate) fn app_config_path(path: &Option<PathBuf>) -> Result<PathBuf> {
+    Ok(path
```
sassman, line 48: this is still a bit of doublication, since the inner returns a result you would not create a new result with `Ok()` but rather relay on the actions that produces a result. For example:

```rust
pub(crate) fn app_config_path(path: &Option<PathBuf>) -> Result<PathBuf> {
    path
        .clone()
        .or_else(|| home::cargo_home().map_or(None, |h| Some(h.join(CONFIG_FILE_NAME))))
        .ok_or_else(|| {
            anyhow!(
                r#"
Unable to resolve path for configuration file.
Use --config option, or place {} in $CARGO_HOME."#,
                CONFIG_FILE_NAME
            )
        })
}
```

sassman: but even better would be to use the [`.with_context`](https://docs.rs/anyhow/1.0.38/anyhow/trait.Context.html) method of anyhow something like:
```rust
pub(crate) fn app_config_path(path: &Option<PathBuf>) -> Result<PathBuf> {
    path
        .clone()
        .or_else(|| home::cargo_home().map_or(None, |h| Some(h.join(CONFIG_FILE_NAME))))
        .with_context(||, format!("Unable to resolve path for configuration file.\nUse --config option, or place {} in $CARGO_HOME.",
                CONFIG_FILE_NAME))
}
```

src/favorites.rs
```diff
@@ -0,0 +1,71 @@
...
+    let favorite_name = args.favorite.as_ref().ok_or_else(|| {
+        anyhow!(r#"Please specify either --git option, or a predefined favorite"#)
+    })?;
+
+    let app_config_path = app_config_path(&args.config)?;
+    let app_config = AppConfig::from_path(app_config_path.as_path())?;
+    let favorite = app_config
+        .favorites
+        .get(favorite_name.as_str())
+        .ok_or_else(|| {
```
sassman, line 54: here too, instead of `.ok_or_else(|| { anyhow!() })` it might be a bit more idiomatic to use `.with_context(|| format!())?`
Also using raw strings is IMO not so common, and I prefer the `format!` macro instead.

src/lib.rs
```diff
@@ -68,25 +80,39 @@ pub struct Args {
...
     create_git(args, &name)?;
-
     Ok(())
 }
 
 fn create_git(args: Args, name: &ProjectName) -> Result<()> {
     let force = args.force;
-    let config = GitConfig::new_abbr(&args.git, args.branch.to_owned())?;
+    let config = GitConfig::new_abbr(
+        &args
```
sassman, line 111: here too, what you find more often is this idiom:

```rust
        &args.git
            .context("Missing option git, or a favorite")?
```

sassman (commented):
few more little nits towards bit more idiomatic rust. Of course this is somehow matter of taste, but I wanted to share my opinion on that.

src/app_config.rs
taurr: It's too bad clippy don't see these :-)

Originally I had it wrapped because I needed the error conversion, that got refactored out but I never caught to rest.
 Didn't know of `with_context`, but hey, can't know of every crate out there - besides, day job is in C#.

I'll fix it!

src/lib.rs
taurr: I'll fix this too

taurr:
@sassman 
I've hopefully fixed all you comments now 😄

sassman (approved):
looks great!

sassman:
Could you please rebase on master?

sassman:
@taurr if you need any hand in resolving the conflicts, just let me know.

taurr:
Rebased my branch on  master!

sassman:
Thanks again that you put in all the efforts in upstreaming this little helpful feature you had already on your fork!

Issue #244, opened 2020-09-12
Add option to disable git repo initialization


avrong:
It can work similarly to `--vcs none` in Cargo in, except here it should be contrasted from `--git` option (describing the template repo url) in `new` (and `init` #193) commands and support only git/no-vcs states in the current implementation.

This would help with rustwasm/wasm-pack#901.

taurr:
#290 disables git initialization when generating stuff within an existing repository - will it be enough, or will an actual option still be needed if/when it is merged?

I'd be happy to volunteer to implement it, though what should the option be called?

sassman:
I guess this was more of a general feature request since cargo new behaves this way. For consistency reasons I think it makes sense.

Of course every contribution is warmly welcomed.

taurr:
@sassman 
Maybe close this one?

sassman:
This will be released with `0.6.0` (pretty soon) and can be tried out on master.

PR #293, opened 2021-02-11
Add --vcs option to avoid initializing git repo

taurr:
#244 asks for an option to specify NOT to initialize the git repository.
This implements such an option, with possibility to be expanded into
other VCS systems.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -102,6 +102,7 @@ use std::{
     collections::HashMap,
     env, fs,
     path::{Path, PathBuf},
+    str::FromStr,
 };
 use structopt::StructOpt;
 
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -71,6 +71,29 @@ pub struct Args {
     /// Enables more verbose output.
     #[structopt(long, short)]
     pub verbose: bool,
+
+    /// Specify the VCS used to initialize the generated template.
+    #[structopt(long, default_value = "git")]
```
sassman, line 76: I've understood the original issue the way that you would not provide a `--vcs git` option but instead use `--git none` option.

sassman: But I'm not 100% sure about it.

taurr: hmmm, this is how I understood it - besides, `--git` is already taken for specifying the git repo for the template :-/

I selected to implement it this way as it also reminds of how it's done for `cargo` itself!

```shell
cargo-init 
Create a new cargo package in an existing directory

USAGE:
    cargo init [OPTIONS] [--] [path]

OPTIONS:
    -q, --quiet                  No output printed to stdout
        --registry <REGISTRY>    Registry to use
        --vcs <VCS>              Initialize a new repository for the given version control system (git, hg, pijul, or
                                 fossil) or do not initialize any version control at all (none), overriding a global
                                 configuration. [possible values: git, hg, pijul, fossil, none]
```

sassman: probably you are right, consistency with cargo-init is what most users would naturally expect

src/lib.rs
```diff
@@ -71,6 +71,29 @@ pub struct Args {
...
+    pub vcs: VCS,
+}
+
+//
+#[derive(Debug, StructOpt)]
+pub enum VCS {
```
sassman, line 82: Just a little nit, could you please rename this enum to the regular [type conventions](https://rust-lang.github.io/api-guidelines/naming.html) that is `UpperCamelCase` so `Vcs`

sassman (changes requested):
found one more small inconsistency, and probably rebasing is also required.

taurr:
Fixed and rebased :-)

sassman:
one more rebase here needed :(

taurr:
Sure - I already anticipated that as I didn't know the order of pulls.
Had to go fetch my daughter from daycare first though.

1 feature rebased, again 😄

sassman (approved):

sassman:
Awesome work @taurr thanks again!

Issue #78, opened 2018-07-31
Project examples as templates


dcarosone:
In #47, one of the possibilities noted is 
> sub dir in a repo

I wanted to make a suggestion that's a little more extensive than this, and I think it's enough of an expansion to list as a separate issue.

A number of projects have `examples/` directories with cargo support for running them directly, and there's also a proposal to let cargo support larger per-directory examples in a sub-structure under there. These are often good starting points for customising your own code; I know I've done this a number of times, especially with some embedded projects. That sounds a lot like a template!

I think is worth combining these proposals and having some specific conventions to make it easier, such that project example directories can always/automatically be templates. I'd like it to be sufficiently low-friction that neither general crate authors nor template consumers need to do anything much extra or unusual to use examples as templates, and offering templates becomes the norm.

This probably involves some changes on both sides, perhaps something like this:
 * example directories look like templates, with placeholders, and have an `example.toml` with placeholder values. Cargo knows how to use this when running in-place as an example.
 * this in turn suggests that making a new example (directory) should itself make a bare template (`cargo new --example`?) so that there's a cohesive story from start to, well, new start :)
 * when given a repo, `cargo generate` can look in the examples directories to find templates, and (in interactive mode) offer a list to pick from, so that repos can be discoverable as either single templates at the root, or more structured collections.

One of the strongest reasons for this is to keep examples and templates in sync with the libraries they represent; cargo should test each of the templates-as-examples just like today. If they're in a separate repo, there's always a risk they'll languish. If they're in the one repo, examples can also be updated to match library code changes on branches.

dcarosone:
> a proposal to let cargo support larger per-directory examples in a sub-structure

I couldn't recall where I'd seen this to link at the time, but it just came up again: 
https://internals.rust-lang.org/t/pre-rfc-project-based-examples-for-cargo-projects/8088

taurr:
What's needed in order to close this?

sassman:
I wanted to start with an example folder to demonstrate certain use cases in order to provide some simple starting points for template authors.

taurr:
@sassman, I just added a folder `example-templates` to the root in #445, as I believe this will be the best way of showing/documenting that feature. If you decide to let #446 go through, then if a template contains `cargo-generate.tom`, users end up being able to expand simply by:
```sh
cargo generate --git <repo>
```

no matter where in the repo the templates are.

[...]

Issue #291, opened 2021-02-08
[RFC] generating partial templates into an existing project


sassman:
**As a** Developer
**I want** to generate a partial into an existing project
**So that** I can apply feature by feature to an existing project

## An example

Let's assume I've started off by a project template a while ago, the project meanwhile grew. 
Not I need the `feature` of a github workflow build pipeline in my existing project.

Let's assume further I have already a project template or a gist of my personal default github workflow pipeline that contains project specific placeholders, like some boolean flags to support building a binary and a library building out of one file. Also it may contain the crate / binary name variables. 

When I run:

```sh
cargo generate --partial my-github-handle/rust-build-pipeline .github/workflows/build.yml
```

I would trigger basically the same project as using cargo-generate the usual way. Just that:

- I can checkout a subset (here `.github/workflows/build.yml`) 
- of a given template repository (here ` my-github-handle/rust-build-pipeline`) 
- into my `$CWD` (here it would expand to `$CWD/.github/workflows/build.yml`)
- optionally I could specify a different target directory (for example `--into src/my/`)

Of course an interactive template rendering is happening as usual.

## Other Use-Cases

the above described behaviour would also enable a some more use-cases. 
For instance one could build a composable project library of `partials` that can be applied step by step. All those components could live in one template repository in purpose specific folders.

taurr:
To be more consistent, and supporting the new `favorite` feature, maybe the syntax should be more in line with:

```shell
cargo generate --git my-github-handle/rust-build-pipeline --partial .github/workflows/build.yml
```

or with favorites:

```shell
cargo generate my-github --partial .github/workflows/build.yml
```

**However**... I am a bit conflicted as to what I think of the feature as a whole - somhow it is as if it is a simple job for a script, not really a feature?

something in the lines of:

```shell
GITPATH=$1
PATTERN=$2

TDIR=$(mktemp -d)
cargo generate --git $GITPATH --name "$TDIR/generated" --force
cp "$TDIR/generated/$PATTERN" .
```

sassman:
It is quite usual that option arguments are not depending on a positional order (think the order cannot be enforced at all)

> However... I am a bit conflicted as to what I think of the feature as a whole - somhow it is as if it is a simple job for a script, not really a feature?

Well, one could argue that the whole feature set of of cargo-generate can be a script. 

IMO: it's all about lifting obstacles, so if I can apply certain template partials to an existing project without scripting it, it would reduce my mental load and helps me a great deal.

[...]

PR #295, opened 2021-02-15
Add feature to use folders as templates (or sub-folders in git).

taurr:
**July 26, 2021 - UPDATE**
Redesigned the feature to only allow selection into git-repositories, thereby avoiding most if not all the problems mentioned in this PR (see #379) .

Using folders alone allows to use templates that are not in git,
but together with git, it allows to specify a sub-folder within
the repository, that contains the actual template.

This allows us to have multiple templates in a single repository.

Relates to #78 (Project examples as templates)

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,8 @@ cargo install cargo-generate --features vendored-openssl
 
 ## Usage
 
+### Template in a git repository
+
 Standard usage is to pass a `--git` flag to `cargo generate` or short `cargo gen`. This will prompt you to enter the name of your project.
 
 ```sh
@@ -52,9 +54,25 @@ You can also pass the name of your project to the tool using the `--name` or `-n
 cargo generate --git https://github.com/githubusername/mytemplate.git --name myproject
 ```
 
-## Favorites
+### Template in a local directory, or git sub-directory
 
-Favorite templates can be defined in a config file, that by default is placed at `$CARGO_HOME/cargo-generate`. 
+Besides `--git` to specify the template repository, `cargo-generate` supports a `--dir` option to specify a local directory as the template.
+
+```sh
+cargo generate --dir ~/my-template-directory --name myproject
+```
+
+Using `--dir` and `--git` together, allows `cargo-generate` to use a sub-directory inside the git repository as the template.
+
+This enables a git repository to contain multiple templates.
+
+```sh
+cargo generate --git username/multi-template-repo --dir directory-in-repo
+```
+
+### Favorites
+
+Favorite templates can be defined in a config file, that by default is placed at `$CARGO_HOME/cargo-generate`.
 To specify an alternative configuration file, use the `--config <config-file>` option.
 
 Each favorite template is specified in its own section, e.g.:
@@ -64,9 +82,12 @@ Each favorite template is specified in its own section, e.g.:
 description = "Demo template for cargo-generate"
 git = "https://github.com/ashleygwilliams/wasm-pack-template"
 branch = "master"
+dir = "directory-for-the-template"
 ```
 
-Both `branch` and `description` are optional, and the branch may be overridden by specifying `--branch <branch>` on the command line.
+`branch`, `description` and `dir` are all optional.
+
+`branch` and `dir` may both be overridden by specifying the corresponding options on the command line.
 
 When favorites are available, they can be generated simply by invoking:
 
@@ -90,7 +111,7 @@ supported placeholders are:
 - `{{crate_name}}`: the snake_case_version of `project-name`
 - `{{os-arch}}`: contains the current operating system and architecture ex: `linux-x86_64`
 
-Additionally all filters and tags of the liquid template language are supported. 
+Additionally all filters and tags of the liquid template language are supported.
 For more information, check out the [Liquid Documentation on `Tags` and `Filters`][liquid].
 
 [liquid]: https://shopify.github.io/liquid
@@ -106,11 +127,11 @@ If you have a great template that you'd like to feature here, please [file an is
 
 ## Template defined placeholders
 
-Sometimes templates need to make decisions. For example one might want to conditionally include some code or not. 
+Sometimes templates need to make decisions. For example one might want to conditionally include some code or not.
 Another use case might be that the user of a template should be able to choose out of provided options in an interactive way.
 Also it might be helpful to offer a reasonable default value that the user just simply can use.
 
-Since version [0.6.0](https://github.com/cargo-generate/cargo-generate/releases/tag/v0.6.0) it is possible to use placeholders in a `cargo-generate.toml` that is in the root folder of a template.  
+Since version [0.6.0](https://github.com/cargo-generate/cargo-generate/releases/tag/v0.6.0) it is possible to use placeholders in a `cargo-generate.toml` that is in the root folder of a template.
 Here [an example](https://github.com/sassman/hermit-template-rs):
 
 ```toml
@@ -150,6 +171,7 @@ fn main() {
 ```
 
 > Note: similar to `dependencies` in the `Cargo.toml` file you can also list them as one liners:
+
 ```toml
 [placeholders]
 hypervisor = { type = "string", prompt = "What hypervisor to use?", choices = ["uhyve", "qemu"], default = "qemu" }
@@ -170,7 +192,7 @@ A placeholder can be of type `string` or `bool`. Boolean types are usually helpf
 
 ### `choices` property (optional)
 
-A placeholder can come with a list of choices that the user can choose from. 
+A placeholder can come with a list of choices that the user can choose from.
 It's further also validated at the time when a user generates a project from a template.
 
 ```toml
@@ -188,12 +210,13 @@ default = 'qemu'
 
 ### `regex` property (optional)
 
-A `regex` property is a string, that can be used to enforce a certain validation rule. The input dialog will keep repeating 
+A `regex` property is a string, that can be used to enforce a certain validation rule. The input dialog will keep repeating
 until the user entered something that is allowed by this regex.
 
 ### Placeholder Examples
 
 An example with a regex that allows only numbers
+
 ```toml
 [placeholders]
 phone_number = { type = "string", prompt = "What's your phone number?", regex = "[0-9]+" }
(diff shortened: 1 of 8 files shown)


src/favorites.rs
```diff
@@ -35,14 +35,13 @@ pub(crate) fn list_favorites(args: &Args) -> Result<()> {
...
-    if args.git.is_some() {
+    if args.git.is_some() || args.folder.is_some() {
         return Ok(());
     }
 
-    let favorite_name = args
-        .favorite
-        .as_ref()
-        .ok_or_else(|| anyhow!("Please specify either --git option, or a predefined favorite"))?;
+    let favorite_name = args.favorite.as_ref().ok_or_else(|| {
```
sassman, line 42: instead of using `ok_or_else` to return the error, you could use `.context("Please specify either --git, --folder, or a predefined favorite")?` to resolve the result or yield the error

src/lib.rs
```diff
@@ -188,61 +203,106 @@ pub fn generate(mut args: Args) -> Result<()> {
...
                 .bold()
                 .red(),
         ),
     };
 
-    create_git(args, &name)?;
-    Ok(())
-}
+    let project_dir = create_project_dir(&project_name, args.force).ok_or_else(|| {
+        anyhow::anyhow!(
```
sassman, line 220: same as above here, it's much more idiomatic to use `.context()?` instead of nesting like this

sassman:
could you also please add a section in the `README.md` that would illustrate this idea a bit?

src/favorites.rs
taurr: I just keep forgetting 😆

src/favorites.rs
```diff
@@ -35,14 +35,14 @@ pub(crate) fn list_favorites(args: &Args) -> Result<()> {
...
-    if args.git.is_some() {
+    if args.git.is_some() || args.folder.is_some() {
         return Ok(());
     }
 
     let favorite_name = args
         .favorite
         .as_ref()
-        .ok_or_else(|| anyhow!("Please specify either --git option, or a predefined favorite"))?;
+        .with_context(|| "Please specify either --git, --folder, or a predefined favorite")?;
```
sassman, line 45: `.with_context` is only needed when you want to format the message with some variables where you would normally use `format!` but then use `.with_context` instead. (reason: the callback is only evaluated when the error appears, so the formatting of the string has only costs when the error case happens)

But here you don't format anything, so what you would use in this case is simply only `.context("Please specify either --git, --folder, or a predefined favorite")` As you can see there is no closure argument, just the plain message.

src/fsutil.rs
```diff
@@ -0,0 +1,15 @@
+use std::{fs, io, path::Path};
+
+pub(crate) fn copy_dir_all(src: impl AsRef<Path>, dst: impl AsRef<Path>) -> io::Result<()> {
+    fs::create_dir_all(&dst)?;
+    for entry in fs::read_dir(src)? {
+        let entry = entry?;
+        let ty = entry.file_type()?;
+        if ty.is_dir() {
```
sassman, line 8: have you tried what happens with symlinks that are pointing to a parent folder?

sassman (commented):
WIP: review feedback

src/fsutil.rs
taurr: It'll fail during copy - `is_dir()` is mutually exclusive with `is_file` and `is_symbolic`, so it'll attempt to copy it as a file.

`fs::copy` fails when the source is not a file...

taurr: So symbolic links inside the template is not currently supported

src/favorites.rs
taurr: Accepted 😄

src/lib.rs
```diff
@@ -129,6 +139,10 @@ pub struct Args {
     /// relative path and only try a GitHub URL if the local path doesn't exist.
     #[structopt(short, long)]
     pub git: Option<String>,
+    /// Folder to copy the template from, or in case the --git option has been specified,
+    /// the sub-folder within the git repository from which to copy the template.
+    #[structopt(long)]
+    pub folder: Option<PathBuf>,
```
sassman, line 145: could you please rename `folder` to `dir`? 
IMO This small change would improve the typing a bit, since `--dir` is as along as `--git`

sassman (changes requested):
I did also some final testing, and I ran into issues when using it inside of an existing git repo.
To give you the example:

```sh
❯ ./target/debug/cargo-generate generate --folder ~/workspaces/rust/hermitcore/hermit-template-rs --name barbak
🔧   Creating project called `barbak`...
🤷  Want to enable network? [true, false] [default: true]: true
🤷  What hypervisor to use? [uhyve, qemu] [default: qemu]: qemu
🤷  Which network to use (only qemu)? [rtl8139, virtio] [default: rtl8139]: rtl8139
Error: Couldn't init new repository

Caused by:
    failed to make directory '/Users/d34dl0ck/workspaces/rust/cargo-generate/barbak/.git': directory exists; class=Filesystem (30); code=Exists (-4)
```

Could you please have a look if that works for you?

src/lib.rs
taurr: I did find 1 issue that might have been the cause. At least it is the only one I could find: I forgot to exclude any `.git` folder when copying over files.

That has now been fixed.

Also renamed `folder` to `dir` as requested.

I actually found it confusing when people started calling directories for folders (wonder where that came up, because AFAIK it used to be only directory or file).
Now I don't know when to call it a folder or a directory - as Walter (Jeff Dunham) would say: pisses me off!

taurr:
@sassman: Ready for next round :-)

Hadn't noticed I had a conflict in 1 file - thank you for not notifying me GitHub!

sassman:
@taurr nice, thanks a lot. Will do today. BTW: did you know that `--git ~/my-cool-template` already works if the template directory is a git repo (local or bare)?

README.md
```diff
@@ -52,9 +54,25 @@ You can also pass the name of your project to the tool using the `--name` or `-n
...
-## Favorites
+### Template in a local directory, or git sub-directory
+
+Besides `--git` to specify the template repository, `cargo-generate` supports a `--dir` option to specify a local directory as the template.
+
+```sh
+cargo generate --dir ~/my-template-directory --name myproject
+```
+
+Using `--dir` and `--git` together, allows `cargo-generate` to use a sub-directory inside the git repository as the template.
```
sassman, line 65: This is somehow really confusing to me. When I omit `--git` and have only `--dir` it is all good and clear. Makes total sense. But when I use `--git` and `--dir` I'm super puzzled. Nothing makes sense anymore in my head. 

I get the idea, and we've discussed the behaviour of `--pick` on #291, so that in general I'd say let's implement that behaviour there, and let's not reuse `--dir` for this 2nd different use case.

My reasoning is: if I'm confused by the duality of `--dir` when used in combination, than there is at least one person out there getting confused too.

src/fsutil.rs
```diff
@@ -0,0 +1,17 @@
...
+            if entry.file_name() != ".git" {
```
sassman, line 9: In fact that excludes the `.git` directory, but what if there are `.svn` or other things inside. I believe I need to land #245 soon, so that we don't have to exclude anything at all.

taurr:
> @taurr nice, thanks a lot. Will do today. BTW: did you know that `--git ~/my-cool-template` already works if the template directory is a git repo (local or bare)?

Yep, It's been working for a while 👍🏼 - but with this one you can use templates outside the git root, or even outside git entirely 😃

sassman (commented):
The main feature looks good, however the dualism of `--dir` used together with `--git` is something I would like to simplify as mentioned below.

src/fsutil.rs
taurr: Sure, but allowing `.git` has a serious risc of trashing an existing repo into which we're generating the template.

Haven't tried this, but using `--git`, which initializes the destination as a new repository, aren't the `.git` folder in the actual template also ignored there? I believe it is.

taurr: Sure enough, generating from a git repository runs `remove_history()` on the cloned repository - this in fact just deletes the `.git` folder too.

taurr: @sassman Given my latest push goes through on this, I've made `--git` and `--dir` mutually exclusive!
I've also introduced `--pick` in order to clear up the functionality (and pave the way for #291)

README.md
taurr: Your reasoning here is sound - and the reason I got confused in #291 - Thus I agree.

Got to admit though, the functionality was right there!!! So instead of simply disabling it I added the `--pick` to clear up the usage pattern. Call it a 2 for 1.

sassman (commented):
the `--dir` feature works great now, so that is approved.

However, the `--pick` needs a bit more love, for the reason below. So depends on you, if you wanna land `--dir` here, then pls out comment the `pick` structops field and let's work on pick separately. 
Or if you want to land both then we need to iterate on it here. Up to you. 

The example where `--pick` behaves unexpected is when a target name, is not fitting into a valid crate:

```sh
❯ ../target/release/cargo-generate gen --dir ~/workspaces/rust/templates/cli-template-tmp --pick .ci --name .ci
⚠️   Renaming project called `.ci` to `ci`...
🔧   Creating project called `ci`...
✨   Done! New project created /Users/d34dl0ck/workspaces/rust/cargo-generate/fooa2/ci
```

In this case the renaming kicks in. 
I wonder if it would not be better to skip the need for `--name` in case for `--pick` so that one can pick one to one. 
And if `--name` is provided we would allow it to be whatever the user enters. 

Thoughts?

taurr:
I'm not sure I follow it entirely.

In the case you describe, I would expect it to create a new dir (specified as `.ci` but auto-renamed to `ci`), in which to expand the template - only you've specified to only pick the `.ci` directory/folder from the template. Thus I'd expect it to create that inside the newly created `ci` dir.

In case you wanted to expand `.ci` directly where you are, we'd need to use the new (and not yet created) `--init` option, which explicitly disables the creation of the first dir (the one renamed to `ci`) and any kind of repo initialization..... I could create that option here too (it's rather trivial with the refactored code-structure)

The way I implemented this is as a simple filter on what gets picked from the template. I've not changed anything on how it behaves when trying to expand the template - which is to always create a dir with the project name, then copy the template contents there and initializing that folder as a new repo (is appropriate). My understanding was this exact behavior is what `--init` should change.

I'm just not sure whether this is the behavior you found and thought surprising?

sassman:
Sorry for the confusion, I'll write the behavior as complementary information on the other issue to draw the picture even better.

taurr:
@sassman: I suggest we scrap this one in favor of #379

PR #296, opened 2021-02-16
WIP: Allow multiple template_values files & use in favorites

taurr:
Allows to specify --template-values-file multiple times, each file being merged in order.

Allows the favorite feature to specify a template-value-file, that is used for getting "default" values.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -93,7 +93,7 @@ mod template;
 mod template_variables;
 
 use anyhow::{Context, Result};
-use config::{Config, ConfigValues, CONFIG_FILE_NAME};
+use config::{Config, CONFIG_FILE_NAME};
 use console::style;
 use favorites::{list_favorites, resolve_favorite};
 use std::{
@@ -104,8 +104,8 @@ use std::{
 };
 use structopt::StructOpt;
 
-use crate::git::GitConfig;
 use crate::template_variables::{CrateType, ProjectName};
+use crate::{config::ConfigValues, git::GitConfig};
 
 #[derive(StructOpt)]
 #[structopt(bin_name = "cargo")]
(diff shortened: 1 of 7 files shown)


sassman:
could you please add at least a test and some documentation on the `README.md` to illustrate how this might be useful?

taurr:
I will add a test and expand on the README.

Actually, this should probably be a draft PR for now - as it seems you may be interested in the folder stuff, I think I need to ponder a little more on this feature... got some nice goodies and usecases for what it enables, though I don't actually think I need to allow multiple option on the command line for it (just ocurred to me!)

sassman:
Sure I will give it some thoughts. Meanwhile please mark the pr as draft. So it can't be merged.

src/lib.rs
```diff
@@ -207,11 +207,23 @@ pub fn generate(mut args: Args) -> Result<()> {
 
 fn create_git(args: Args, name: &ProjectName) -> Result<()> {
     let force = args.force;
+    use itertools::*;
     let template_values = args
         .template_values_file
-        .as_ref()
-        .map(|p| Path::new(p))
-        .map_or(Ok(Default::default()), |path| get_config_file_values(path))?;
+        .iter()
```
sassman, line 213: the `create_git` function has grew a lot, could you please refactor this part that does now the template value collection into an own function something like `collect_template_values_from_files` to reduce the responsibilities of this function. 

You could then also add a very small unit test for the function.

sassman (changes requested):
In general I find the idea appealing, so I'm very glad if you could finish the missing pieces that have been addressed.

sassman:
@taurr this needs a rebase

taurr:
yeah, I know - kinda more focused on #295 right now and TBH. I don't want to redo everything once that hopefully lands.

This feature may end up being easier for me to just re-implement once we complete #295 - there's a reason for this one being draft for now :-)

That'll teach me not to attempt multiple features at the same time! (or, one could hope, but it's most likely not going to change anything)

taurr:
Rebased, but failing SSH tests in workflow

taurr:
Closed as this needs reimplementation anyway.

Issue #389 created to discuss/agree upon the scope.

PR #309, opened 2021-03-07
feat(favorites:fallback): backwards compatibility

sassman:
this PR ensures backwards compatibility to v0.5.x so that the favorite feature does not create any harm for long term users. So in detail that looks like this:

 - when a user provides an argument such as `sassman/hermit-template-rs` without `--git`
 - then this is looked up in the favourites but in case it is missing
   we log a warning and use it as a git repo as if the user provided `--git sassman/hermit-template-rs`
 - so we remain backwards compatible
 - a user will likely notices about the favorites feature (warning message)
 - insteaf of failing handle what the user want more gracefully, more peace of mind

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -243,12 +243,6 @@ dependencies = [
 
 [[package]]
 name = "cfg-if"
-version = "0.1.10"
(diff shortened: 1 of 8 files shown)


sassman:
FYI @taurr

PR #312, opened 2021-03-09
Add rust-arduino-generator template

HKalbasi:


taurr:
Created issue HKalbasi/rust-arduino-generator#2 in response to this PR.

Now that we're deprecating `TEMPLATES.md", this really shouldn't be needed.

PR #328, opened 2021-04-24
bugfix(git): Add support for ssh urls

### LINKED ISSUE #325: Git SSH URL Support
The [Github Docs](https://docs.github.com/en/enterprise-server@3.0/github/getting-started-with-github/about-remote-repositories#about-remote-repositories) specify that a remote repository URL must have one of the following forms

- https://[hostname]/user/repo.git
- git@[hostname]:user/repo.git

As a result, if cargo-generate handles the former, I would expect it to handle the latter.

```
cameron@chowder ~> cargo generate --version
cargo-generate 0.6.1
cameron@chowder ~> cargo generate --git https://github.com/rustwasm/wasm-pack-template.git --name my-project
🔧   Creating project called `my-project`...
     Running `git fetch --force --update-head-ok 'https://github.com/rustwasm/wasm-pack-template.git' 'refs/heads/master:refs/remotes/origin/master'`
✨   Done! New project created /home/cameron/my-project
cameron@chowder ~> cargo generate --git git@github.com/rustwasm/wasm-pack-template.git --name my-project-2
🔧   Creating project called `my-project-2`...
Error: ⛔   Git Error: unexpected http status code: 404; class=Http (34)
cameron@chowder ~ [1]> cargo generate --git ssh://git@github.com/rustwasm/wasm-pack-template.git --name my-project-2
🔧   Creating project called `my-project-2`...
Error: ⛔   Git Error: authentication required but no callback set; class=Ssh (23)
```

If this is a features maintainers would like to have, then I would be happy to implement it.

SkamDart:
Resolves #325


sassman:
@SkamDart Thanks for your PR! 

I'm going to have a deeper look this coming weekend.

src/git.rs
```diff
@@ -65,7 +64,37 @@ impl GitConfig {
...
+    callbacks.credentials(|_url, username_from_url, _allowed_types| {
+        Cred::ssh_key(
+            username_from_url.unwrap_or("git"),
+            None,
+            Path::new(&format!("{}/.ssh/id_rsa", env::var("HOME").unwrap())),
+            None,
+        )
+    });
+
+    let branch = match args.branch {
```
sassman, line 79: thought: this code, specifically the fallback for the default branch name to `"master"` differs a bit bit from the code that is used in case of a http url. (see Line 119 - 123). It would be very good to reuse this code as much as possible to have a sane, consistent behaviour for both sort of urls.

sassman: suggestion: I think it would be simpler to have just one function for the `create` as it was before, and do a little pre ceremony for the credentials in case the url is a ssh url. That way you would basically depend literally on the very same code and just to the things that are needed for the other url type upfront and then the rest should not differ at all.

sassman (changes requested):
First and foremost, thanks a lot for your PR. I appreciate every contribution. 
Please find some little remarks below in the code review that I would appreciate if you can take the time and check them out. They should not be too much work though.

src/git.rs
```diff
@@ -65,7 +64,37 @@ impl GitConfig {
...
 
-pub(crate) fn create(project_dir: &Path, args: GitConfig) -> Result<String> {
+fn create_with_ssh(project_dir: &Path, args: GitConfig) -> Result<String> {
+    let mut callbacks = RemoteCallbacks::new();
+
+    callbacks.credentials(|_url, username_from_url, _allowed_types| {
+        Cred::ssh_key(
+            username_from_url.unwrap_or("git"),
+            None,
+            Path::new(&format!("{}/.ssh/id_rsa", env::var("HOME").unwrap())),
```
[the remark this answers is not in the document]
SkamDart: Yup totally agree this is not the right way to handle it.

Totally fine making the change though there are a possible options here.

If you're okay with adding a dependency, it seems like the supported cross-platform way to do this is through the [dirs](https://docs.rs/dirs/3.0.2/dirs/index.html) crate.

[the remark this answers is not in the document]
sassman: The dependency tree that comes with `dirs` is not small but acceptable. IMO we can leverage from it on other places too. So go for it 😃

src/git.rs
```diff
@@ -65,10 +64,22 @@ impl GitConfig {
...
+        dirs::home_dir().ok_or_else(|| git2::Error::from_str("$HOME was not set"))?;
+    priv_key.push(".ssh/id_rsa");
+    Cred::ssh_key(username_from_url.unwrap_or("git"), None, &priv_key, None)
+}
+
 pub(crate) fn create(project_dir: &Path, args: GitConfig) -> Result<String> {
     let temp = Builder::new().prefix(project_dir).tempdir()?;
     let config = Config::default()?;
     let remote = GitRemote::new(&args.remote);
+    let is_http = args.remote.to_string().contains("http");
```
sassman, line 82: suggestion: to harden that code a bit more, I'd suggest to check if the url starts with that protocol, since contains can yield false positives, like a repo that has `http` in it's name like `git@github.com:hyperium/http.git`

src/git.rs
```diff
@@ -86,8 +97,19 @@ pub(crate) fn create(project_dir: &Path, args: GitConfig) -> Result<String> {
...
             let repo = git2::Repository::init(&temp.path())?;
             let mut git_remote = repo.remote_anonymous(remote.url().as_str())?;
-            git_remote.connect(git2::Direction::Fetch)?;
-            let default_branch = git_remote.default_branch()?;
+
+            let default_branch = if is_http {
+                git_remote.connect(git2::Direction::Fetch)?;
+                git_remote.default_branch()?
+            } else {
+                let mut cb = RemoteCallbacks::new();
```
sassman, line 105: question: is that not a bit of code duplication to like 135?

SkamDart: `git2::RemoteCallbacks` doesn't implement `Clone` or `Copy` moved the creation to a function that returns a RemoteCallback instead.

src/git.rs
SkamDart: Good point. Switched to checking for `ssh://`

src/git.rs
```diff
@@ -65,10 +64,24 @@ impl GitConfig {
     }
 }
 
+fn git_credentials_callback<'a>() -> Result<RemoteCallbacks<'a>> {
+    let mut priv_key =
+        dirs::home_dir().ok_or_else(|| git2::Error::from_str("$HOME was not set"))?;
```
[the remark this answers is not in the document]
SkamDart: Interesting. TIL there is a `anyhow::Context` implementation for `Option<T>`

sassman (changes requested):
could you please have a look at the comments and then it would be very great if you could prepare the test case as explained [above](https://github.com/cargo-generate/cargo-generate/pull/328#pullrequestreview-654732180)

sassman:
/to @SkamDart 

not sure if you have seen it already, I made feat comments and wanted to ask you if you need any further input or help?

SkamDart:
@sassman not an abandoned pr

Will be taking a look again this weekend.

sassman:
Hey @SkamDart

I wanted to check back if you need any help with this branch? Is there anything I can support you with?

Greetings

sassman:
thanks for your work, I've adopted and finalised it in #372

Issue #211, opened 2020-02-02
Generate one project from many templates. 


xoac:
I was looking for a nice tool that will help me start with project. This looks quite good but I would propose to modify idea a little. This could make a steps that generate a full project instead of using just one source. For example:

1. Add license (choose one or many)
1. Add CI (this would be a template that will ask some question and create a basic defaults). 
1. Add keepchangelog
1. Add code-of-conduct
1. Setup it as CLI / Webassebly or no_std project.

This idea is to make generate very customable. Instead of using one source allow using many of them. Of course there can conflict - but this would be small disadvantage of this solution. 

At the end of day people would be creating own templates in seconds by removing or adding new templates.  

This can start with:
```
cargo generate 
 Past link to source of 1 template:
>  https://git..... (here a link to it)
     > Question from this templates
     > (..)

 Past link to source of 2 template or empty to end:
> 
(...)
```

Threre could be a configuration file in `~/.config/cargo-generate/templates` where we could store own steps:
```
cli:
 - http://git...
 - http://git..
```
Usage
```
cargo generate cli 
```

k0pernicus:
Hi @xoac 

Based on my experience, best would be a shell script with a basic for-loop in your URL instead of letting the CLI doing this...
Especially for error handling and user interaction.

Something like this could be helpful for you, no?

```sh
for URL in https://github.com/x/project-a https://github.com/x/project-b; do
    cargo generate --git $URL ...
done 
```

sassman:
relates to #291, so that partials can be used to slice and dice a project into several repositories as well. 
Since this is a very specific use case I guess we won't implement a generic "pass as many templates" to compile one project out of it. 

I recommend a shell function to do so.

Issue #47, opened 2018-07-25
RFC: What can be a template?


killercup:
## Context

When executing cargo-generate a user needs to pass an identifier for the template they want to generate a project from.

## Questions

What can be such template identifiers?

How can we extract a cargo-generate compatible template from them?

## Possibles solutions

One or more of

- git repo
- directory
- crate
- sub dir in a repo

durka:
If downloading from Github is going to be common, it seems like a good idea to support `username/repo` syntax instead of the full URL.

Also, what about gists?

mmstick:
No reason not to support all the possible solutions.

anirudhb:
I think if it has a Cargo.toml, it is a template.

durka:
So I could then just cargo-generate any crate, and it would download it and
replace the author's name with mine? That seems kinda weird...

anirudhb:
Not like that! 😅
I was thinking if it _wasn't_ published to crates.io, then we could do that...

killercup:
> I think if it has a Cargo.toml, it is a template.

People want to use this for mdbook repos too: #14

rahul-thakoor:
Yes, in this case having a book.toml makes it a template. 

Are we able to pass in local git repos? This could be helpful when creating and testing templates.

softprops:
Porteurbars went with local directly , GitHub, and raw gitburls https://github.com/softprops/porteurbars/blob/master/src/git.rs#L10

Local dirs are especially useful for local fast testing of a new template

skade:
I would like to register interest in "subdirectory of a repos". Our use case is managing a number of assignments in one git respos and we would like to ship starter templates along with them in subdirectories.

taurr:
How is "crate as a template" supposed to work?
Do we need it?

sassman:
One way I imagine it to work is downloading from crates.io and the. Treating it similar as a local `--path` template folder.

Not sure if we need it, because I believe that a crate that is only a template might even not compile and there for might even be not publishable.

I would say we threat this issue as solved unless some folks have a real use case for crate as template.

taurr:
>Not sure if we need it, because I believe that a crate that is only a template might even not compile and therefore might even be not publishable.

>I would say we threat this issue as solved unless some folks have a real use case for crate as template.

Wholeheartedly agreed!

PR #379, opened 2021-07-25
Feature: specify subdir in repo as template

taurr:
This change allows to specify a sub-folder inside a git repository for use as the template.

By doing so, it allows us to e.g. generate projects from examples inside a repo, or to have a repo that contains multiple templates.

Relates to #47 and #78 

This pull-request will also make #295 obsolete, as there is no need to specify folders as templates, and thus we avoid a lot of the problems specified in #295 and #291, allowing for easier implementation of #291 (and thus #211).

@sassman Sorry for the long timeout here - as it seems we get no feedback for #295, I suggest this one instead. I still like the idea of `--pick` for #291, but it kinda conflicted and this is my proposal to remedy that!

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,16 @@ You can also pass the name of your project to the tool using the `--name` or `-n
 cargo generate --git https://github.com/githubusername/mytemplate.git --name myproject
 ```
 
+If the git repository contains multiple templates, the specific subfolder in the git repository may be specified like this:
+
+```sh
+cargo generate --git "https://github.com/githubusername/mytemplate.git <relative-template-path>"
+```
+
+> NOTE: The specified `relative-template-path` will be used as the actual template root, whether or not this is actually true!
+
+> NOTE: When using the `subfolder` feature, `cargo-generate` will search for
+
 ## git over ssh
 
 New in version [0.7.0] is the support for both public and private and ssh git remote urls.
@@ -82,7 +92,7 @@ cargo generate --git rustwasm/wasm-pack-template --name mywasm
 
 ## http(s) proxy
 
-New in version [0.7.0] is automatic proxy usage. So, if http(s)_PROXY env variables are provided, they 
+New in version [0.7.0] is automatic proxy usage. So, if http(s)\_PROXY env variables are provided, they 
 will be used for cloning a http(s) template repository. 
 
 ## Favorites
@@ -94,12 +104,13 @@ Each favorite template is specified in its own section, e.g.:
 
 ```toml
 [favorites.demo]
-description = "Demo template for cargo-generate"
+description = "<optional description, visible with --list-favorites>"
 git = "https://github.com/ashleygwilliams/wasm-pack-template"
-branch = "master"
+branch = "<optional-branch>"
+subfolder = "<optional-subfolder>"
 ```
 
-Both `branch` and `description` are optional, and the branch may be overridden by specifying `--branch <branch>` on the command line.
+Values may be overridden using the CLI arguments of the same names (e.g. `--subfolder` for the `subfolder` value).
 
 When favorites are available, they can be generated simply by invoking:
 
@@ -110,7 +121,7 @@ cargo gen <favorite>
 or slightly more involved:
 
 ```cli
-cargo generate demo --branch master --name expanded_demo
+cargo generate demo --branch mybranch --name expanded_demo --subfolder myfolder
 ```
 
 > NOTE: when `<favorite>` is not defined in the config file, it is interpreted as a git repo like as if `--git <favorite>`
@@ -294,6 +305,8 @@ include = ["Cargo.toml"]
 # include and exclude are exclusive, if both appear we will use include
 exclude = ["*.c"]
 ```
+
+The `cargo-generate.toml` file should be placed in the root of the template. If using the `subfolder` feature, the root is the `subfolder` inside the repository, though `cargo-generate` will look for the file in all parent folders until it reaches the repository root.
 
 ## Cargo gen - alias
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -51,15 +51,21 @@ impl TryFrom<String> for Config {
 }
 
 impl Config {
-    pub(crate) fn new<P: AsRef<Path>>(path: P) -> Result<Option<Self>> {
-        match fs::read_to_string(path) {
-            Ok(contents) => Config::try_from(contents)
-                .map(Option::from)
-                .map_err(|e| e.into()),
-            Err(e) => match e.kind() {
-                ErrorKind::NotFound => Ok(None),
-                _ => anyhow::bail!(e),
+    pub(crate) fn from_path<P>(path: &Option<P>) -> Result<Option<Self>>
+    where
+        P: AsRef<Path>,
+    {
+        match path {
+            Some(path) => match fs::read_to_string(path) {
+                Ok(contents) => Config::try_from(contents)
+                    .map(Option::from)
+                    .map_err(|e| e.into()),
+                Err(e) => match e.kind() {
+                    ErrorKind::NotFound => Ok(None),
+                    _ => anyhow::bail!(e),
+                },
             },
+            None => Ok(None),
         }
     }
 }
@@ -89,7 +95,7 @@ mod tests {
         )
         .unwrap();
 
-        let config = Config::new(&config_path).unwrap().unwrap();
+        let config = Config::from_path(&Some(config_path)).unwrap().unwrap();
 
         assert_eq!(
             config.template,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -97,6 +97,7 @@ use config::{Config, ConfigValues, CONFIG_FILE_NAME};
 use console::style;
 use favorites::{list_favorites, resolve_favorite_args};
 use std::{
+    borrow::Borrow,
     collections::HashMap,
     env, fs,
     path::{Path, PathBuf},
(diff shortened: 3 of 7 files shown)


taurr:
@sassman After the changes to 0.7.0 with SSH, I can't run tests neither locally nor here on Github - any ideas?

sassman:
Locally should work. What's the error message?

The github action right now failed because of access to a secret variable.. I will fix it ASAP

taurr:
test basics::ssh_remote::it_should_support_a_public_repo ... FAILED
test basics::ssh_remote::it_should_support_a_private_repo ... FAILED

failures:

---- basics::ssh_remote::it_should_support_a_public_repo stdout ----
thread 'basics::ssh_remote::it_should_support_a_public_repo' panicked at 'Unexpected failure.
code-1
stderr=```"Error: ⛔   Git Error: Failed to retrieve list of SSH authentication methods: Failed getting response; class=Ssh (23)\n"```
command=`"/home/johnny/projects/github/cargo-generate/target/debug/cargo-generate" "generate" "--git" "git@github.com:ashleygwilliams/wasm-pack-template.git" "--name" "foobar-project"`
code=1
stdout=```"🔧   Creating project called `foobar-project`...\n💡   Using private key: `$HOME/.ssh/id_rsa` for git-ssh checkout\n"```
stderr=```"Error: ⛔   Git Error: Failed to retrieve list of SSH authentication methods: Failed getting response; class=Ssh (23)\n"```
', /home/johnny/.cargo/registry/src/github.com-1ecc6299db9ec823/assert_cmd-1.0.7/src/assert.rs:155:13

---- basics::ssh_remote::it_should_support_a_private_repo stdout ----
thread 'basics::ssh_remote::it_should_support_a_private_repo' panicked at 'Unexpected failure.
code-1
stderr=```"Error: ⛔   Git Error: Failed to retrieve list of SSH authentication methods: Failed getting response; class=Ssh (23)\n"```
command=`"/home/johnny/projects/github/cargo-generate/target/debug/cargo-generate" "generate" "--git" "git@github.com:cargo-generate/wasm-pack-template.git" "--name" "foobar-project"`
code=1
stdout=```"🔧   Creating project called `foobar-project`...\n💡   Using private key: `$HOME/.ssh/id_rsa` for git-ssh checkout\n"```
stderr=```"Error: ⛔   Git Error: Failed to retrieve list of SSH authentication methods: Failed getting response; class=Ssh (23)\n"```
', /home/johnny/.cargo/registry/src/github.com-1ecc6299db9ec823/assert_cmd-1.0.7/src/assert.rs:155:13


failures:
    basics::ssh_remote::it_should_support_a_private_repo
    basics::ssh_remote::it_should_support_a_public_repo

test result: FAILED. 37 passed; 2 failed; 1 ignored; 0 measured; 0 filtered out; finished in 10.91s

error: test failed, to rerun pass '--test integration'

README.md
```diff
@@ -62,6 +62,14 @@ You can also pass the name of your project to the tool using the `--name` or `-n
 cargo generate --git https://github.com/githubusername/mytemplate.git --name myproject
 ```
 
+If the git repository contains multiple templates, the specific subfolder in the git repository may be specified like this:
+
+```sh
+cargo generate --git "https://github.com/githubusername/mytemplate.git#<relative-template-path>"
```
sassman, line 68: note: TBH I'm not yet fully convinced by this "special" URL notation. 

I'm in favour of keep the git URL a regular URL and better add an unnamed argument to keep the folder in there separated.

I know the that this would conflict with the <favorite> argument but this can be solved by a bit of logic:
1. <favorite> is only allowed in disjunction with `-g | --git`
2. so if <favourite> is present then <relative-template-path> would be the 2nd unnamed argument and optional
3. if `-g | --git` is present then the first named argument is taken as <relative-template-path> and not as it is now as <favourite>

src/config.rs
```diff
@@ -51,7 +51,16 @@ impl TryFrom<String> for Config {
 }
 
 impl Config {
-    pub(crate) fn new<P: AsRef<Path>>(path: P) -> Result<Option<Self>> {
+    pub(crate) fn from_path<P: AsRef<Path>>(path: P) -> Result<Option<Self>> {
+        let path = path.as_ref();
+
+        if !path.exists() {
```
sassman, line 57: question: can you please elaborate why this block there with going to parent folders is required?

sassman (commented):
please find my questions and notes in the review

README.md
taurr: Hmm, this might actually be doable - so, if `-g` | `--git` argument is specified, we use the unnamed argument (normally used to specify a 'favorite') as the internal folder? I like that 😃 

In this case, when a folder needs to be specified in a favorite, we need to name the argument though... any idea?

src/config.rs
taurr: If the folder specified contains the template config it should be used from there, but in case it doesn't its very nice that it picks up the config in the parent folder.

Somewhat akin to how it works with git config files etc.

It can be argued though, that the folder traversal could/should stop at the git repository root path.

README.md
sassman: hm, good point. I forgot about the configuration file structure and that the `<subfolder>` would make sense too.

I suggest to add a named optional parameter in the favourites config with name `subfolder`.

like:

```toml
[favorites.cli]
description = "CLI template for a rust binary"
git = "https://github.com/sassman/cli-template-rs.git"
branch = "master"
subfolder = "examples/demo"
```

so this `subfolder = "examples/demo"` is then the new part.

src/config.rs
sassman: I see that this might be very useful. So I suggest:

1. this smart config finding logic is documented
2. we need here some mechanism to prevent a traversal beyond the actual template folder, so as you say stopping at the template root

README.md
taurr: Regarding name resolution, if the user does not specify a `--name`, should we use the subfolder name (if specified) instead of the repo name?

I'm thinking yes!

taurr:
Famous words: It works on my machine(s) 😄

src/lib.rs
```diff
@@ -204,72 +213,168 @@ pub fn generate(mut args: Args) -> Result<()> {
...
+    match &args.subfolder {
+        Some(subfolder) => {
+            println!(
+                "{} {} `{}`{}",
+                emoji::WRENCH,
+                style("Using template subfolder").bold(),
+                style(&subfolder).bold().yellow(),
+                style("...").bold()
+            );
+            template_base_dir.path().join(subfolder)
```
sassman, line 292: issue: a check if that final folder is really a folder would be good here. If you pass a file instead of a folder the following error is happening and that should be handled gracefully. 

```sh
cargo run -- generate demo --name foobar28 README.md
    Finished dev [unoptimized + debuginfo] target(s) in 0.12s
     Running `target/debug/cargo-generate generate demo --name foobar28 README.md`
🔧   Creating project called `foobar28`...
🔧   Using template subfolder `README.md`...
[1/1]   Done:
Error: Not a directory (os error 20)

```

tests/integration/favorites.rs
```diff
@@ -51,57 +51,60 @@ branch = "{branch}"
...
-        .contains(r#"description = "git-template""#));
+        .failure();
 }
 
 #[test]
-fn it_can_use_favorites() {
-    let favorite_template = create_template("favorite-template");
-    let (_config, config_path) = create_favorite_config("test", &favorite_template);
-    let dir = tmp_dir().build();
+fn favorite_with_subfolder() -> anyhow::Result<()> {
```
sassman, line 75: question: could you also add a test for the not happy path, e.g. passing an invalid subfolder or a file to the `<subfolder>` argument?

src/lib.rs
sassman: additionally we need to ensure here, that this final folder is really part of the `template_base_dir`, I was able to exploit the mechanism by handing an absolute path to the `<subfolder>` argument. This needs to be prevented.

you can test it by doing this:

```sh
mkdir /tmp/foo-bar
echo "Hello {{project-name}}" >> /tmp/foo-bar/hex.md

cargo run -- generate demo --name foobar26 /tmp/foo-bar
    Finished dev [unoptimized + debuginfo] target(s) in 0.14s
     Running `target/debug/cargo-generate generate demo --name foobar26 /tmp/foo-bar`
🔧   Creating project called `foobar26`...
🔧   Using template subfolder `/tmp/foo-bar`...
[1/1]   Done: hex.md
✨   Done! New project created /Users/d34dl0ck/workspaces/rust/cargo-generate/foobar26

# cargo-generate has used a folder not part of the project template
cat foobar26/hex.md
Hello foobar26
```

sassman (changes requested):
looks very good, unfortunately I found 2 blockers related to some checks we currently miss on the subfolder. You can find all details in the review.

src/lib.rs
```diff
@@ -204,72 +213,168 @@ pub fn generate(mut args: Args) -> Result<()> {
...
+                crate::emoji::WARN,
+                style("Symbolic links not supported").bold().red(),
+            )
+        }
     }
     Ok(())
 }
 
+fn locate_template_file<N, T>(
+    name: &N,
```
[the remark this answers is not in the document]
taurr: Cause I'm human and make errors 😆

README.md
```diff
@@ -322,5 +348,5 @@ license, shall be dual licensed as above, without any additional terms or
 conditions.
 If you want to contribute to `cargo-generate`, please read our [CONTRIBUTING notes].
 
-[CONTRIBUTING notes]: CONTRIBUTING.md
-[0.7.0]: https://github.com/cargo-generate/cargo-generate/releases/tag/v0.7.0
\ No newline at end of file
+[contributing notes]: CONTRIBUTING.md
```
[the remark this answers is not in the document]
taurr: LOL... I totally agree here, and finally got my act together to tell vscode NOT to do any auto-formatting nor any removal of trailing spaces in *.md files.

taurr:
Hmm, I probably shouldn't have force pushed my latest commit with the review changes - but I thought: hey @sassman is keen on having a clean git history, so I'll make it easy...

And now I don't know how to get Github to recognize I've made the needed changes 😦

sassman:
I recommend squashing at the end of the review when all things are settled.

sassman (approved):

sassman:
good job and many thanks @taurr

PR #381, opened 2021-07-26
feat(templace:choices): better ux for template choice parameters

sassman:
- templates can contain paramters with a list of options and a default
- this is now handled with the dialogouer capabilities for select from a list
- users can now choose via keys and don't need manual typing anymore


sassman:
image

Issue #159, opened 2019-05-28
Can dir names and file names be templated?


kitsuneninetails:
Can I have a file like `test_{{project-name}}.rs` and have that convert during generation to `test_my-project.rs`?

If not, that's something that probably needs to be added IMHO to give this any real utility over and above a simple `find . -type f -exec sed -ie "s/{{project-name}}/$1/g" {} \;`

Changing dir names and file names based off templates is much harder and a more interesting/helpful problem to solve IMHO.  It'd be nice if this could do that.

k0pernicus:
I think we can include this feature, as the program iterates on each file of the template project.
We just need to include a check on each file entry [here](https://github.com/ashleygwilliams/cargo-generate/blob/54a85ed98ee4ad06c8228dd60ac070ff483291dd/src/template.rs#L102).

k0pernicus:
@kitsuneninetails I am writing a feature request for your demand - I wrote the code and the tests and both work.
I will clean the code, and try a PR soon.

xortive:
https://github.com/ashleygwilliams/cargo-generate/pull/161 was going to solve this for 0.4.0, but it's being pushed back to a later release. https://github.com/ashleygwilliams/cargo-generate/pull/182 has some more details on the motivation behind this.

jmjoy:
Need this feature too!

Michael-F-Bryan:
I can think of a couple use cases where I'd like this, for example creating a template for a workspace where there's a `{{project_name}}` crate at the project root.

Is there anything I can do to help with the implementation or testing?

k0pernicus:
Hi @Michael-F-Bryan , you can take a look at how I proposed it a loooong time ago (https://github.com/ashleygwilliams/cargo-generate/pull/161) and propose a solution to solve the issue from @xortive at this time (see #182 for the motivation behind the revert at this time).

taurr:
#255 also relates to this!

notdanilo:
Is there any progress being made for this issue?

csarn:
I am also interested in this feature. Python's "cookiecutter" does that, too.

sassman:
solved with #386 and released in v0.8.0

PR #386, opened 2021-07-28
feat(template:filenames): template filenames now

sassman:
- solves #159 and is partially based on PR #174
- addresses some issues from PR #182

  > what to do about characters coming out of a template tag and being invalid when part of a filename

  - now filenames are sanitized after templating

  > what to do about using template tags in filenames along with the include option in cargo-generate.toml when you can't specify a filename with curly braces in it ergonomically using glob syntax

  - simply not supported, hence exclusion must be configured with the filename before substitution.

- add tests to check for exploitation
- README.md updated

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -215,6 +215,7 @@ dependencies = [
  "predicates",
  "regex",
  "remove_dir_all 0.7.0",
+ "sanitize-filename",
  "serde",
  "structopt",
  "tempfile",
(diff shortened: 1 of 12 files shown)


sassman:
@k0pernicus catching up on this that you have once done. Maybe you remember further problems that we better test here to avoid side effects?

taurr (approved):
Looks nice.

Hadn't considered that most filesystems (if not all) will allow the `{{variable}}` filenaming - but it's great that it does. Makes it very easy to understand.

Issue #387, opened 2021-07-29
Allow template from a directory instead of a git repo


NOBLES5E:

sassman:
do you mean to support something like this:

```sh
# a local template
mkdir /tmp/my-template
echo "hello {{project-name}}" >> /tmp/my-template/README.md


cargo generate /tmp/my-template --name my-project-from-a-local-template-folder

# all has been generated
cat my-project-from-a-local-template-folder/README.md
hello my-project-from-a-local-template-folder
```

NOBLES5E:
Exactly! I didn't notice that the folder can be passed directly as an argument. Thanks!

NOBLES5E:
I tried the example you gave but found the following error:

```
⚠️   Favorite /tmp/my-template not found in config, using it as a git repo url
```

sassman:
It's not yet supported. 

But iirc we had once a small prototype that did the trick. So eventually this feature will land it short time.

NOBLES5E:
glad to hear that

sassman:
Will be released with the next version.

PR #390, opened 2021-07-30
Allow template to be a plain folder.

taurr:
If template fails to be cloned as a git repo (due to not being a repo!), try to just copy it as a plain folder.

This allows us to use plain folders as templates, and will make it easier when developing new templates.

Relates to #387


sassman (approved):
Good work!

Fixes #387

Issue #389, opened 2021-07-30
FEA: specify `template_values_file` in multiple places


taurr:
This issue takes over from an old attempt of my own (#296), and here's the idea...

What I'd like to end up with is:

1. Be able to specify default `template_values_file` in the global `.cargo-generate.toml` file
    * Maybe specify values directly in a `[values]` section?
2. Override values with `template_values_file`, definable for each defined favorite.
    * Maybe specify values directly in a `[favorites.<name>.values]` section?
3. Override with `template_values_file` from the CLI.
    * Maybe add support for `CARGO_GENERATE_<name>` style environment variables here?
    * Maybe add `--define`|`-d` as CLI argument to specify individual values?

In my mind, this should make the "value" system more complete, in that we can specify them at several well defined layers.

Before charging ahead though, I'd like to know if there's any comments/ideas/objections?

@sassman any good comments - knowing the history of #296 which basically is a failed and less well defined attempt at this?

sassman:
> 1. Be able to specify default template_values_file in the global .cargo-generate.toml file
>     - Maybe specify values directly in a [values] section?

Could you elaborate how this would look like?

--

> Maybe specify values directly in a [favorites.<name>.values] section?

This would open up some flexibility and is a nice idea.

--

> Maybe add `--define|-d` as CLI argument to specify individual values?

This is a very good idea! 
I would even dedicate an own PR for it. It would add lot's of flexibility to users, and is very easy to use. IIRC folks have even asked for this somewhere (cannot find the issue now).

taurr:
Example global `.cargo-generate.toml` file for the possibilities above...

```toml
[default]
# In case values are not specified anywhere else, take 'em from here (or ask interactively)
template_values_file = "~/my_default_template_values.toml"

# Specifying default values directly in the settings file
[default.values]
network_enabled = "true"
another_value = "just a demo"

[favorites.demo]
description = "Demo template for cargo-generate"
git = "https://github.com/ashleygwilliams/wasm-pack-template"
branch = "main"
# For this favorite, override the default values with the ones defined here (relative to the template folder)
template_values_file = "demo_template_values.toml"

# overriding values, from favorites
[favorites.demo.values]
another_value = "lets change the default value here!"
new_value = "and even add a new one"
```

> NOTE: This file is already used for specifying the favorites, and may be selected using the `--config` option.

[...]

PR #392, opened 2021-07-30
Allow `--define` to specify values

taurr:
Relates to #389.

Just couldn't help myself 😄

### DIFF
--- a/src/template_variables/mod.rs
+++ b/src/template_variables/mod.rs
@@ -8,7 +8,7 @@ use crate::{config::ConfigValues, emoji, Args};
 use anyhow::Result;
 use console::style;
 use regex::Regex;
-use std::{collections::HashMap, fs, path::Path};
+use std::{collections::HashMap, fmt::Display, fs, path::Path};
 use toml::Value;
 
 pub(crate) use authors::{get_authors, Authors};
@@ -25,7 +25,7 @@ pub(crate) fn resolve_template_values(args: &Args) -> Result<HashMap<String, Val
             read_template_values_file(path)
         })?;
 
-    add_cli_defined_values(&mut template_values, args)?;
+    add_cli_defined_values(&mut template_values, &args.define)?;
 
     Ok(template_values)
 }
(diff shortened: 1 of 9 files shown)


sassman (changes requested):
It's a very good start!

Could you please add a test case that would define a variable of a choice parameter with a value that is not part of the choices? 

Same goes for boolean values that are no actual booleans.

taurr:
Hmmm, just noticed - there's not even any tests for the `--template-values-file` flag 😮 

Will create one tomorrow then!

tests/integration/values.rs
```diff
@@ -0,0 +1,155 @@
...
+
+#[test]
+fn it_accepts_template_values_file() {
+    let template = tmp_dir()
+        .file(
+            "my-values.toml",
+            indoc! {r#"
+                [values]
+                my_value = "content of my-value"
+            "#},
```
sassman, line 17: I think this indentation level here is not correct. It should be on the same as in the lines above right?

tests/integration/values.rs
```diff
@@ -0,0 +1,155 @@
...
+        )
+        .file(
+            "Cargo.toml",
+            indoc! {r#"
+                [package]
+                name = "{{project-name}}"
+                description = "{{my_value}}"
+                version = "0.1.0"
+            "#},
```
sassman, line 26: same as above

tests/integration/values.rs
```diff
@@ -0,0 +1,155 @@
...
```
sassman, line 132: as above

tests/integration/values.rs
```diff
@@ -0,0 +1,155 @@
...
+
+#[test]
+fn cli_value_overrides_others() {
+    let template = tmp_dir()
+        .file(
+            "my-values.toml",
+            indoc! {r#"
+                [values]
+                my_value = "content of file-value"
+            "#},
```
sassman, line 123: as above

sassman (commented):
looks very good. Just found a few little nits.

tests/integration/values.rs
taurr: I guess thats a matter of taste - personally I like it this way, as it's all the same parameter for the `indoc!` macro, separate from the filename parameter preceeding it. The `indoc!` macro is nice because it removes all the extra fluff spacing thats there just to make it easier to scan through the code.

I'd prefer to keep it this way if possible.

tests/integration/values.rs
taurr: same 🤷🏽

tests/integration/values.rs
taurr: same

tests/integration/values.rs
taurr: same

src/template_variables/mod.rs
```diff
@@ -3,7 +3,68 @@ mod crate_type;
...
+            "{} {} {}",
+            emoji::ERROR,
+            style("Values File Error:").bold().red(),
+            style(e).bold().red(),
+        ),
+    }
+}
+
+fn add_cli_defined_values(template_values: &mut HashMap<String, Value>, args: &Args) -> Result<()> {
+    let regex = Regex::new(r"(\w+)=(.+)").unwrap();
```
sassman, line 48: should be better allow a more advanced identifier on the left hand side? 

e.g. something like `project-description` is not covered at the moment.

sassman (approved):

tests/integration/values.rs
sassman: sure, fine with me

src/template_variables/mod.rs
taurr: it is now - as well as a few unittests ✔️

taurr:
> It's a very good start!
> 
> Could you please add a test case that would define a variable of a choice parameter with a value that is not part of the choices?
> 
> Same goes for boolean values that are no actual booleans.

Regarding the tests for values of the wrong type - I feel like we should still have them, but probably in another PR, as this one doesn't mess with that behavior.

Should I uncover errors (I don't even know what the expected outcome is right now), it'll affect the `--template-values-file` flag and maybe interactive mode too.

sassman (approved):
nice

PR #395, opened 2021-08-04
filenames: allow directory to contain template tag

dave-tucker:
This commit allows for directories to also contain template tags

The use case for this would be creating a template that contained multiple crates, where the directory name for those crates could be derived from the template.


sassman (approved):
lgtm, thanks for your contribution

Issue #396, opened 2021-08-04
using pattern in dir doesn't work correctly


dave-tucker:
my pr #395  doesn't actually work.

for one, I forget to `mkdir -p` before write, so there are issues with the dir not being created.
also, when I've generated using a test project, I end up with duplicate directories.
e.g `{{author}}/main.rs` and `foo/main.rs`

i'll look in to this, add an integration test and open a pr to fix.
sorry!

PR #397, opened 2021-08-04
template: Process Directories and Files

dave-tucker:
In order to properly allow for templates in directory names, we must
adjust the WalkDir to include both files and directories AND to yield
contents first.

Given that the project name is bar, and the directory tree:

| {{project-name}}/foo.rs
| {{projcet-name}}

- {{project-name}}/foo.rs is read and the contents rendered
- fs:create_dir_all will create the bar directory
- fs::write will write bar/foo.rs
- the {{project-name}} directory is then deleted with fs::remove_dir_all

Which will provide:

| bar/foo.rs

Fixes: #396

Command output from this commit:
```
🔧   Creating project called `foo`...
[ 1/12]   Done: .appveyor.yml
[ 2/12]   Done: .gitignore
[ 3/12]   Done: .travis.yml
[ 4/12]   Done: Cargo.toml
[ 5/12]   Done: LICENSE_APACHE
[ 6/12]   Done: LICENSE_MIT
[ 7/12]   Done: README.md
[ 8/12]   Done: src/lib.rs
[ 9/12]   Done: src/utils.rs
[10/12]   Done: src
[11/12]   Done: tests/web.rs
[12/12]   Done: tests
✨   Done! New project created /home/dave/dev/cargo-generate/foo
```

### DIFF
--- a/src/template.rs
+++ b/src/template.rs
@@ -168,10 +168,11 @@ pub(crate) fn walk_dir(
 
     let files = WalkDir::new(project_dir)
         .sort_by_file_name()
+        .contents_first(true)
         .into_iter()
         .filter_map(Result::ok)
-        .filter(|e| !e.file_type().is_dir())
         .filter(|e| !is_git_metadata(e))
+        .filter(|e| e.path() != project_dir)
         .collect::<Vec<_>>();
     let total = files.len().to_string();
     for (progress, entry) in files.into_iter().enumerate() {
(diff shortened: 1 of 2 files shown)


src/template.rs
```diff
@@ -168,9 +168,9 @@ pub(crate) fn walk_dir(
 
     let files = WalkDir::new(project_dir)
         .sort_by_file_name()
+        .contents_first(true)
         .into_iter()
         .filter_map(Result::ok)
-        .filter(|e| !e.file_type().is_dir())
         .filter(|e| !is_git_metadata(e))
```
sassman, line 174:
```suggestion
        .filter(|e| !is_git_metadata(e))
        .filter(|e| e.path() != project_dir)
```

sassman: this is the change for the last unnecessary substitution on the root level

sassman:
thanks for your fix!

> One odd side effect is that I can't figure out how to prune the root directory from the WalkDir:

Above you'll find a solution for the issue

dave-tucker:
Thanks @sassman :pray: I've incorporated your suggested fix and it works as expected!

sassman (approved):

sassman:
Thanks a lot @dave-tucker

PR #398, opened 2021-08-06
Only use cross for linux prebuilt binaries

jashandeep-sohi:
@sassman, I'm proposing a few changes to the prebuilt binaries workflow:

- Only use cross for linux builds. It's not really needed for macos & windows. On windows it also causes binaries to be built without the .exe extension.
- Try running running the built binary as a sanity check. Should we also be running tests for each of the targets too?
- Lock down the release action to a specific version & set `omitPrereleaseDuringUpdate` to avoid resetting the pre-release state of a release.


sassman:
> Should we also be running tests for each of the targets too?

Since this is triggerd on release creation, and this therefore is happening only after a successful "release" action run where tests etc. are already considered, I would not opt for testing at this stage again.

Running the binary as you've already done is good enough IMO.

sassman (approved):

PR #404, opened 2021-08-10
Specify values via env vars.

taurr:
Relates to #389

- Accepts env var `CARGO_GENERATE_TEMPLATE_VALUES_FILE` to specify the `template-values-file` option.
- Accept env vars on the form `CARGO_GENERATE_VALUE_<key>`, to set individual values (akin to `--define`)

> NOTE: Windows does **not** support passing in values for lowercase variables via the environment variables. Windows insist on using uppercase for all ENV vars.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,13 +205,13 @@ Since version [0.6.0](https://github.com/cargo-generate/cargo-generate/releases/
 Here [an example](https://github.com/sassman/hermit-template-rs):
 
 ```toml
-[placeholders.hypervisor]
+[placeholders.HYPERVISOR]
 type = "string"
 prompt = "What hypervisor to use?"
 choices = ["uhyve", "qemu"]
 default = "qemu"
 
-[placeholders.network_enabled]
+[placeholders.NETWORK_ENABLED]
 type = "bool"
 prompt = "Want to enable network?"
 default = true
@@ -219,10 +219,10 @@ default = true
 
 As you can see the `placeholders` configuration section accepts a table of keywords that will become the placeholder name.
 
-In this example the placeholder `hypervisor` and `network_enabled` will become template variables and can be used like this:
+In this example the placeholder `HYPERVISOR` and `NETWORK_ENABLED` will become template variables and can be used like this:
 
 ```rs
-{% if network_enabled %}
+{% if NETWORK_ENABLED %}
 use std::net::TcpListener;
 
 fn main() {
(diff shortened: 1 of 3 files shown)


tests/integration/values.rs
```diff
@@ -7,22 +7,33 @@ use assert_cmd::prelude::*;
...
 
 #[test]
-fn it_accepts_template_values_file() {
+fn it_accepts_template_values_from_multiple_places() {
     let template = tmp_dir()
+        .file(
+            "my-env-values.toml",
+            indoc! {r#"
+                [values]
+                MY_VALUE1 = "env-file-value"
```
sassman, line 16: question: could you also devote a dedicated section in the README.md to document this behaviour?

especially important for template provider to highlight is the fact that the variable names should be UPPER CASED to be able to be feed via env variables.

tests/integration/values.rs
```diff
@@ -34,16 +45,29 @@ fn it_accepts_template_values_file() {
...
         .arg("--template-values-file")
         .arg(template.path().join("my-values.toml"))
-        .arg(template.path())
         .current_dir(&dir.path())
+        .env(
+            "CARGO_GENERATE_TEMPLATE_VALUES_FILE",
+            template.path().join("my-env-values.toml"),
+        )
+        .env("CARGO_GENERATE_VALUE_MY_VALUE2", "env-def-value")
+        .env("CARGO_GENERATE_VALUE_MY_VALUE3", "env-def-value")
```
sassman, line 60: the test illustrates it all, but I'd rather split it up for 2 reasons:
1. if it fails it would be very hard to spot where in the whole mechanics it failed
2. the mental load to oversee the entire mechanics in this test is not quite small

sassman (changes requested):
LGTM. 

Only the documentation part is missing then we are good to go.

README.md
```diff
@@ -205,24 +205,24 @@ Since version [0.6.0](https://github.com/cargo-generate/cargo-generate/releases/
 Here [an example](https://github.com/sassman/hermit-template-rs):
 
 ```toml
-[placeholders.hypervisor]
+[placeholders.HYPERVISOR]
```
sassman, line 208: I feel not very contempt about this upper casing, but I had an idea how we can avoid it.

What if we would provide like for the templating context always the following 3 identical versions of a variable:

- named as originally provided (e.g. via `--define`)
- lower cased 
- upper cased

by doing so we would prevent the situation to have upper cased variables in the template but still we could provide by the environment variables that are upper cased.

sassman: but is the problem not that a template author would like to work with template variables that then would work for all users on all platforms equally?

In order to guarantee this, we could also set out for the convention that template variables should always follow a lower case naming conventions, and we make sure that variable names provided via env variables are lower cased to fulfil the convention. In all other cases they are as users provide them. No further magic.

taurr: > In order to guarantee this, we could also set out for the convention that template variables should always follow a lower case naming conventions, and we make sure that variable names provided via env variables are lower cased to fulfil the convention. In all other cases they are as users provide them. No further magic.

This I like :-)

A simple convention, a small fix to the implementation and mentioning it in the README... consider this one done!

PR #409, opened 2021-08-13
Allow specification of default template values

### LINKED ISSUE #46: RFC: Sources of values
e.g. "looking at cli params, env vars, global config files, or, in the end, by prompting the user"

### LINKED ISSUE #389: FEA: specify `template_values_file` in multiple places

taurr:
Fixes #389

Seems to also close #46

### DIFF
--- a/README.md
+++ b/README.md
@@ -350,7 +350,7 @@ Default values may be specified in the config file (specified with the `--config
 placeholder1 = "default value"
 
 [favorites.my_favorite]
-git = some template git repo
+git = https://github.com/githubusername/mytemplate.git
 
 [favorites.my_favorite.values]
 placeholder1 = "default value overriding the default"
--- a/README.md
+++ b/README.md
@@ -335,9 +335,27 @@ cargo generate template-above
 
 #### Template values file via environment variable
 
-As a last resort, the user may use the environment variable `CARGO_GENERATE_TEMPLATE_VALUES` to specify a file with default values.
+The user may use the environment variable `CARGO_GENERATE_TEMPLATE_VALUES` to specify a file with default values.
 
 For the file format, see [above](#valuesfile)
+
+#### Default values
+
+Default values may be specified in the config file (specified with the `--config` flag, or in the default config file `$CARGO_HOME/cargo-generate`)
+
+**Example config file:**
+
+```toml
+[values]
+placeholder1 = "default value"
+
+[favorites.my_favorite]
+git = some template git repo
+
+[favorites.my_favorite.values]
+placeholder1 = "default value overriding the default"
+placeholder2 = "default value for favorite"
+```
 
 ## Include / Exclude
 
--- a/README.md
+++ b/README.md
@@ -350,7 +350,7 @@ Default values may be specified in the config file (specified with the `--config
 placeholder1 = "default value"
 
 [favorites.my_favorite]
-git = https://github.com/githubusername/mytemplate.git
+git = "https://github.com/githubusername/mytemplate.git"
 
 [favorites.my_favorite.values]
 placeholder1 = "default value overriding the default"
--- a/src/favorites.rs
+++ b/src/favorites.rs
@@ -1,6 +1,8 @@
+use std::collections::HashMap;
+
 use crate::{
     app_config::{AppConfig, FavoriteConfig},
-    emoji, info, Args,
+    emoji, warn, Args,
 };
 use anyhow::{anyhow, Result};
 use console::style;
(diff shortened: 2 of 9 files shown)


src/favorites.rs
```diff
@@ -42,34 +48,48 @@ pub(crate) fn list_favorites(app_config: &AppConfig, args: &Args) -> Result<()>
...
-    let (git, branch, subfolder) = app_config
+    let (values, git, branch, subfolder) = app_config
         .favorites
-        .get(favorite_name.as_str())
+        .as_ref()
+        .and_then(|f| f.get(favorite_name.as_str()))
         .map_or_else(
             || {
-                info!(
+                warn!(
```
sassman, line 71: question: I wonder why that is now a warning? I think a favourite config is still optional, hence I would stick with an info message.

src/git.rs
```diff
@@ -231,7 +231,7 @@ fn git_clone_all(project_dir: &Path, args: GitConfig) -> Result<String> {
                 return Err(e.into());
             }
 
-            info!("Template does not seem to be a git repository, using as a plain folder");
+            warn!("Template does not seem to be a git repository, using as a plain folder");
```
sassman, line 234: note: there the warning is reasonable and fine with me

tests/integration/favorites.rs
```diff
@@ -7,7 +7,7 @@ use assert_cmd::prelude::*;
 use indoc::indoc;
 use std::path::PathBuf;
 
-fn create_template(name: &str) -> Project {
+fn create_template(description: &str) -> Project {
```
sassman, line 10: suggestion: I found it more reasonable to name those parameters along their meaning or name within the template, here we have the `project-name` not the `description`, I suggest to stick with the naming

README.md
```diff
@@ -335,10 +335,28 @@ cargo generate template-above
...
+Default values may be specified in the config file (specified with the `--config` flag, or in the default config file `$CARGO_HOME/cargo-generate`)
+
+**Example config file:**
+
+```toml
+[values]
+placeholder1 = "default value"
+
+[favorites.my_favorite]
+git = some template git repo
```
sassman, line 353: suggestion: provide a real value to be more specific, including the right type + syntax

sassman (changes requested):
please find some nits inline, it would be nice if you have a look

src/favorites.rs
taurr: hmm, I just renamed the `info!` macro, as it really was outputting a warning, so I didn't change the meaning/output here.

I could however change it, so we just write it as info ??

tests/integration/favorites.rs
taurr: The argument is actually the description!

`project-name` is not being substituted here, but left for the template engine. `description` however is. Please notice this is inside a `format!` macro, thus the `{{project-name}} is ignored, while the `{}` is not.

src/favorites.rs
sassman: I see, have not recognized it.

tests/integration/favorites.rs
sassman: Thanks for clarifying

README.md
```diff
@@ -335,10 +335,28 @@ cargo generate template-above
...
+Default values may be specified in the config file (specified with the `--config` flag, or in the default config file `$CARGO_HOME/cargo-generate`)
+
+**Example config file:**
+
+```toml
+[values]
+placeholder1 = "default value"
+
+[favorites.my_favorite]
+git = https://github.com/githubusername/mytemplate.git
```
sassman, line 353: I believe this needs to be put in double quotes.

sassman (approved):

README.md
taurr: I believe you are right :-)

taurr:
Fixed and rebased

sassman (approved):

PR #411, opened 2021-08-14
fix(tests:linux): use the current directory for canonicalize result

orhun:
This PR removes the assumption from `should_canonicalize` test that is always being run in a subdirectory which is in `/home`.

For example, while building `cargo-generate` for Arch Linux in a clean chroot, tests are being run in `/build/cargo-generate/src` and this canonicalize test is failing as expected:

```
thread 'git::should_canonicalize' panicked at 'assertion failed: canonicalize_path(&PathBuf::from(\"../\")).unwrap().starts_with(\"/home/\")', src/git.rs:106:5
```

### DIFF
--- a/src/git.rs
+++ b/src/git.rs
@@ -103,9 +103,13 @@ fn should_canonicalize() {
         .unwrap()
         .starts_with("/Users/"));
     #[cfg(target_os = "linux")]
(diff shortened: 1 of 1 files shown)


taurr (approved):
LGTM

sassman (approved):
Nice catch.

PR #417, opened 2021-08-16
build.yml: add publish-dry-run job

MarcoIeni:


.github/workflows/build.yml
```diff
@@ -107,4 +107,18 @@ jobs:
         uses: actions-rs/audit-check@v1
         continue-on-error: true
         with:
-          token: ${{ secrets.GITHUB_TOKEN }}
\ No newline at end of file
+          token: ${{ secrets.GITHUB_TOKEN }}
+          
+  publish-dry-run:
```
sassman, line 112: question: would a dry run with here not cause a failure in cases where the `Cargo.toml` version number was not changed? Which is quite often the case.

MarcoIeni: If the `Cargo.toml` version is the same of the one in crates.io the dry-run is still successful

MarcoIeni: to avoid issues like that you could use https://github.com/sunng87/cargo-release

Issue #400, opened 2021-08-06
Better support for choice lists with many choices


taurr:
Using a `choice` placeholder with a lot of options is currently not feasible, as the number of choices scroll off the screen in the interactive mode and makes it next to impossible to select anything.

We should implement e.g. paging to handle when there are more choices than can naturally fit in the terminal.

It would be super if it also was possible to search in the choice list.

taurr:
As info and testing, try out:

```shell
cargo generate --git taurr/embedded-template-rs application
```

it's my template for a complete embedded application using ferrous tooling, still in the works.

sassman:
I see. But instead of us tackling this issue I'd recommend we upstream that issue to the `dialoguer` crate maintainers, since I cannot imagine we are the only ones interested in a solution for this issue.

I found https://github.com/mitsuhiko/dialoguer/issues/118 that is slightly but none that is directly addressing this very issue.

sassman:
upstream issue is opened

deg0nz:
> It would be super if it also was possible to search in the choice list.

Searching would be possible via recently merged `fuzzy-search` feature, but it currently does not support paging. (Just FYI)

taurr:
Just tried using newest dialoquer (via git dependency), enabling the "fuzzy-select" feature, then in `interactive.rs`:64 changing to use `FuzzySelect::with_theme` instead of `Select::with_theme`

Works like a charm 😄 

Now we're just waiting for a fix to mitsuhiko/dialoguer#125 and a release, then this issue can be closed (with the above corrections).

deg0nz:
I looked into the `cargo-generate` code and was able to identify the error that actually happens here. Please see my last comment in mitsuhiko/dialoguer#125 🙂

sassman:
upstream issue is now solved, `0.9.0` makes an explicit `.paged` call obsolete (in fact it was completely removed). Will provide an update for it.

PR #423, opened 2021-08-19
feat: enable paging for long choice lists

taurr:
while waiting for mitsuhiko/dialoguer#125, manually enable paging in dialoguer.

Related to #400 

Still feel like we should wait until `dialoguer` make their next relase for closing #400 though, as we should enable `fuzzy-search` too (already on their master)


sassman (approved):
yeah, make sense

PR #424, opened 2021-08-22
enumerate list_favorites conflicts

MarcoIeni:
I noticed that those parameters are not used when the `list_favorites` flag is passed, so in theory they should be treated as conflicts, right?

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -28,7 +28,6 @@ pub struct Args {
         conflicts_with = "force",
         conflicts_with = "template_values_file",
         conflicts_with = "silent",
-        conflicts_with = "config",
         conflicts_with = "vcs",
         conflicts_with = "lib",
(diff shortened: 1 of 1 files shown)


src/args.rs
```diff
@@ -18,7 +18,24 @@ pub enum Cli {
...
+        long,
+        conflicts_with = "git",
+        conflicts_with = "subfolder",
+        conflicts_with = "path",
+        conflicts_with = "branch",
+        conflicts_with = "name",
+        conflicts_with = "force",
+        conflicts_with = "template_values_file",
+        conflicts_with = "silent",
+        conflicts_with = "config",
```
sassman, line 31: issue: `config` should not be conflicting since it relates to the favourites files

sassman (commented):
One little nit to investigate

PR #426, opened 2021-08-23
ci: add documentation check

MarcoIeni:
I also:
- removed `--all` which is deprecated, in favor of `--workspace`
- surrounded with `<>` the markdown links, so that they are clickable from the docs
- removed duplicate documentation in main.rs, which is available in lib.rs


sassman (approved):
Lgtm

PR #428, opened 2021-08-23
README: improve cargo_generate_version example

MarcoIeni:
I think that having a version specified as `">=0.8.0"` is more common than `">0.8.0"`.
Maybe in the copy paste, one may forget to add the `=` and this can lead to an error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -391,7 +391,7 @@ Using the supported `cargo-generate.toml` file, the template author may setup ve
 
 ```toml
 [template]
-cargo_generate_version = ">0.8.0"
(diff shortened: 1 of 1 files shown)


sassman (approved):
👍

taurr:
:-) This one is kinda funny, cause the original ">0.8.0" was choosen because it wasn't possible on v0.8.0, and I didn't know what version it would arrive in....

We should actually update the doc to ">=0.9.0", as that's when it became possible :-)

MarcoIeni:
I created https://github.com/cargo-generate/cargo-generate/pull/433 to solve this issue :)

Issue #415, opened 2021-08-15
Deprecate use of `.genignore` file?


taurr:
Can we deprecate the use of `.genignore` now that include/exclude can be done in the `cargo-generate.toml` file? 

Maybe at least in the `README.md`? (Better yet: remove any mention of it)

taurr:
@sassman please close this one - I believe I retired `.genignore` by discouraging its use in https://github.com/cargo-generate/cargo-generate/pull/431

Issue #18, opened 2018-07-10
Pre/Post Hooks


epage:
This is something else that has been handy from [cookiecutter](https://cookiecutter.readthedocs.io/en/latest/advanced/hooks.html) but is more questionable due to security.

Use cases from my python cookiecutter translated to Rust.
- Custom SCM support
- Verifying ident (see #17)
- `cargo update && git add Cargo.lock` for bins

The main question is how would we handle this in Rust?
- shell scripts aren't portable
- `cargo-script`?

To alleviate security concerns, we could have flags for
- Showing what will be done to the system (prints the scripts)
- Runs the template but not the hooks (requiring manual intervention)

MarcoIeni:
I did a little research of what projects allow to run rust code. Here is what I found:
- https://crates.io/crates/rust-script
- https://crates.io/crates/cargo-play
- https://crates.io/crates/runner
- https://crates.io/crates/cargo-eval (fork of cargo-script which wasn't updated in 4 years)

PR #431, opened 2021-08-23
feat: include files conditionally

### LINKED ISSUE #230: Introduce features, a way to include files conditionally
It would be really cool if `cargo-generate` would have something like cargo's features, where a user could select to include certain configuration/setup files of a template if he wants to.

An example: My template contains GitHub Actions workflow files for creating releases on tag pushes on master as well as creating documentation on the `gh-pages` branch when a push to master happens. The template would, by default, include everything. But if a user doesn't want the docs workflow, he could run something similar to this:

```
cargo-generate --git <url> --no-docs-workflow
cargo generate --git <url> --no-default-features --features docs-workflow
```

The first is a somewhat nicer syntax for this use-case I think. The second one would be the same as the one from cargo.

I imagine something like this inside the configuration file:
```toml
[[features]]
name = "docs-workflow"
include = [".github/workflows/docs.yml"]

[[features]]
name = "bors"
include = ["bors.toml"]
```

This comes really handy in situations where templates offer a lot of pre-defined configuration files with sane and good default values for various services (GitHub Actions, bors etc). That way, a user can just pick what he wants and be done.

Feedback is appreciated!

__Edit__: Additionally, this could be expanded upon composing features into "super-features". For example:
```rust
[[features]]
name = "workflows"
uses = ["docs-workflow", "publish-workflow", "pr-workflow"]
```

This would introduce more code to maintain, as something like cyclic dependency checks have to be performed on the features before running them.

taurr:
Allow special syntax in `cargo-generate.toml` in order to allow files to be
included/excluded/ignored based on variables.

Also allow placeholders to be included conditionally.

`features` feature not needed, as this is more flexible, and can be coupled with the `--template-values-file`
or `--define` options in order to create predefined sets of variables.

Closes #230


src/lib.rs
```diff
@@ -199,16 +201,13 @@ fn prepare_local_template(args: &Args) -> Result<(TempDir, PathBuf, String), any
...
-                exclude: _,
-            }),
-        placeholders: _,
-    }) = template_config
+fn check_cargo_generate_version(template_config: &Config) -> Result<(), anyhow::Error> {
+    if let Some(config::TemplateConfig {
+        cargo_generate_version: Some(requirement),
+        include: _,
+        exclude: _,
+        ignore: _,
```
MarcoIeni, line 209:
```suggestion
        ..
```
clippy should complain about this I think

src/lib.rs
```diff
@@ -430,35 +429,97 @@ fn expand_template(
...
-        template_config.and_then(|c| c.template),
-        &mut pbar,
-    )?;
+    match template::walk_dir(dir, template, template_cfg, &mut pbar) {
+        Ok(_) => {}
+        Err(e) => {
+            let e = format! {"{}", e};
+            bail!(e);
+        }
+    }
```
MarcoIeni, line 456: isn't `if let Err(e) = emplate::walk_dir(dir, template, template_cfg, &mut pbar)` better in this case?
Because you are ignoring the `Ok` branch

src/template.rs
```diff
@@ -27,6 +27,80 @@ fn engine() -> liquid::Parser {
...
+        input: &dyn ValueView,
+        _runtime: &dyn Runtime,
+    ) -> Result<liquid_core::model::Value, liquid_core::error::Error> {
+        let input = input
+            .as_scalar()
+            .ok_or_else(|| liquid_core::error::Error::with_msg("String expected"))?;
+        let input = input.into_string().to_string().to_lowercase();
+
+        Ok(liquid_core::model::Value::scalar(
+            if !input.is_empty() && input != "false" && input != "0" {
```
MarcoIeni, line 62: this should be documented in the readme before we ship this

src/lib.rs
taurr: Unfortunately it didn't, and I didn't think of it either... oh well, good thing you noticed @MarcoIeni. 😃 

I'll look into these minor issues tonight - also need to do some writing for the README. This was just a first attempt at solving the problem, so if y'all like the proposal, I count this a huge success.

My own biggest issue is the use of `liquid` for the conditional expressions. It can be a bit cumbersome, but I don't have a better idea for now. The `truthy` and `falsy` filters were an attempt to ease this issue a little, but it might not be enough - `liquid` is really for rendering text and not for evaluating expressions.

src/lib.rs
taurr: I'm at work atm. but if my memory recalls it correctly, this piece might even be there due to temporary debugging needs - obviously needs a little love

src/template.rs
taurr: `README.md` didn't exactly receive any attention yesterday - as noted: this was first draft, to let you all comment on a concrete implementation of @MarcoIeni's idea (with a little extra spice!).

src/lib.rs
MarcoIeni: I don't know liquid well, but `|truthy` and `|falsy` don't look so bad I think :)
Isn't it possible to use the same syntax as `'{% if crate_type == "bin" %}true{% endif %}'`?
In this way we can also have different placeholders based on string comparisons, not only on boolean expressions.

MarcoIeni: By the way if this requires a longer discussion we could just handle "include files conditionally" in this PR and open a separate issue/PR on "include placeholders conditionally" if that's feasible and if you think it's better.

taurr: it is indeed possible - its actually the ONLY way if it wasn't for the 2 filters :-)

taurr: > By the way if this requires a longer discussion we could just handle "include files conditionally" in this PR and open a separate issue/PR on "include placeholders conditionally" if that's feasible and if you think it's better.

I'd actually prefer to just have it all here. The only reason to delay in my mind, is if we'd want something other than `liquid`for the conditional expression syntax - I just don't see why or what.

Regarding the placeholders - I added those as well as the ignore setting, as I really feel like this should all be possible from within `cargo-generate.toml`.

MarcoIeni: > it is indeed possible - its actually the ONLY way if it wasn't for the 2 filters :-)

Oh, so you can use the two filters in addition to that other syntax. I.e. it's just syntax sugar. I see :)

taurr:
@MarcoIeni: I did some cleanup and rebased the feature. Also created a section in the README.

MarcoIeni:
1. Do you want to use Rhai instead of liquid or in addition to it? I.e. all the previous template will continue to work?
2. Do you want to use Rhai for including files conditionally instead of this implementation?

taurr:
> * Do you want to use Rhai instead of liquid or in addition to it? I.e. all the previous template will continue to work?

My thought was only for this conditional stuff. `Liquid` is uniquely positioned in that it is intended for textual rendering, fitting the purpose of a template engine 100%.

> * Do you want to use Rhai for including files conditionally instead of this implementation? If yes, can you provide an example?

Where `liquid` falls short is as a scripting language, and in this case: when you just need to evaluate an expression. What I had to do, was basically declare that any text at all means `true`, no text means `false`... then it occurred to me that rendering no text at all from a language that is supposed to render text can be hard - so I made the exceptions `no`, `none`, `false`. It's all patches on the bigger problem: `liquid` is not intended to solve a simple problem like `networked && chip=="cortex-m4"`.

`Rhai` on the other hand is intended for solving those kinds of questions, and IMHO it would suck for text rendering - and (guessing/looking into the future here) even allows us to define methods that we deem "safe" in the context of a template, thus allowing to define scripts that could be run as "post-hooks" (see #18). 


What I need here is just your opinions before trying to use `Rhai` for THIS issue!
I would rather investigate this now than be stuck with `liquid` for this syntax, if there's a better alternative that fits the problem domain better.

`Liquid` works, but I don't think its a natural nor elegant fit for the problem.

MarcoIeni:
I see, that sounds great! Can you provide an example of using rhai for this use case?

sassman:
that this point I'd like to add my 2 cents, things that I've had in mind for longer already. 

I'm not very happy with liquid, because it's not very flexible and also the graceful handling of missing variables (see #422) was already a super nasty hack.

If we would look at a very popular and flexible alternative we would see that [tera](https://tera.netlify.app/docs/#functions) has the nice features to register user defined functions that are written in rust. Besides the security concerns, this would be a very nice way of allowing template authors to ship with a set of functions that they then could also use in their templates. 

As said from a security point of view this might be a nightmare, but if it comes to flexibility this is for sure far beyond what we could do for every individual template author.

taurr:
@sassman I'm all in favor of changing the template rendering engine to something like `Tera`, and we should probably do it very soon if we're going to do it!

**That said, using [`Rhai`](https://rhai.rs/book/) here doesn't impact the used template rendering engine at all.**

This last commit is me changing to use `Rhai` for evaluating the conditions. It makes reading/writing the conditions **much** easier and natural, while also allowing much more flexibility.

While using `Tera` for the template rendering would allow for a better experience writing the actual template files - and even provide us with a lot more flexibility in the future, it's job is still to render text. Methods provided here are still supposed to support that usecase; rendering text.

`Rhai` allows us the more natural feel when writing these conditions. In the future it could also provide us (or rather the template authors) the capability to write post-template-expansion hook scripts - exactly what the scripts could do depends on the methods we provide the engine with. `Rhai` is all about running/evaluating user provided scripts/expressions, returning e.g. bool. It's **not** for rendering the templates - that would be up to `liquid` or `Tera` (I'd probably also prefer the latter).

Final words: I'd very much like to use e.g. `Tera` also, but as I've tried to illustrate; it's really another issue. One doesn't prevent the other in this case, as they do not conflict in any way.

taurr:
@MarcoIeni just look at the latest commit :-)

MarcoIeni:
taurr, the examples look great!
It's exciting that we could also implement pre/post hooks with rhai!

Regarding tera..running arbitrary code when generating a template sounds a little bit scary ^^' But as taurr said it's a separate issue.

taurr:
> Are we using rhai only for evaluating the cargo-generate.toml? Or is it enabled in the whole template?

It's specific in use for cargo-generate.toml
 
> Regarding tera..running arbitrary code when generating a template sounds a little bit scary ^^' But as taurr said it's a separate issue.

*Defending `Tera` for later use as a rendering engine here...*
I've looked a little at `Tera` too (just some of the docs) - from what I see, you can not invoke arbitrary code with that either. Rather it's just like `Rhai` in that we as tool developers can create a set of methods that we expose for use from `Tera`. In that term it is "safe" for as long the methods we expose don't do any harm! 

The big difference between the 2 is the output (at least as I see it). For `Tera` its text, while for `Rhai` it's more likely the side-effects or the value of some expression (as in our conditions case).

taurr:
Cleaned up the code a little, rebased.

I do feel this is ready for actual review, lets see if the good man @sassman likes this too.

sassman:
great work @taurr. I hope to devote some intense review time this week, but so far looks pretty nice.

taurr:
Closes #415 
As it directs users towards using the mecanisms in `cargo-generate.toml` instead of using `.genignore`

README.md
```diff
@@ -417,6 +438,50 @@ cargo_generate_version = ">=0.9.0"
 
 The format for the version requirement is [documented here](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html).
 
+## Conditional template settings.
+
+Using `cargo-generate.toml`, values and some [`Rhai`](https://rhai.rs/book/) syntax, the template author can make certain conditional decisions before expansion of the template.
+
+`include`, `exclude`, `ignore` and `placeholders` can all be used in sections that are only used based upon the value of one or more values, possibly input by the user using the interactive prompt (if the values in question are defined as placeholders in the non-conditional section).
+
+Example:
```
sassman, line 447: suggestion: in order to make this whole thing a little easier to digest I would not pack such a dense example, and rather go for maybe 2-4 smaller examples, where every single one would focus on certain things. 

Such big examples require a lot of mental load to process and try to illustrate too many points at the same time as a result new users getting more easily lost. 

This is just a suggestion thou.

README.md
```diff
@@ -417,6 +438,50 @@ cargo_generate_version = ">=0.9.0"
...
+[conditional.'crate_type == "lib"']
+ignore = [ "src/main.rs" ]
+# include = [ "..." ]
+# exclude = [ "..." ]
+
+[conditional.'crate_type != "lib"'.placeholders]
+hypervisor = { type = "string", prompt = "What hypervisor to use?", choices = ["uhyve", "qemu"], default = "qemu" }
+network_enabled = { type = "bool", prompt = "Want to enable network?", default = true }
+
+[conditional.'license == "MIT"']
```
sassman, line 468: praise: you've chosen pretty nice examples thou

src/template_filters.rs
```diff
@@ -0,0 +1,260 @@
+use anyhow::Result;
```
sassman, line 1: question: could we split this file into an own folder as a module `template_filter/` where we could have for every filter a very small own file?

README.md
taurr: Thus, I've chosen to keep the example, but tried to improve upon the explanation :-)

PR #437, opened 2021-08-25
refactor: add more clippy warnings

taurr:
Went through the list of clippy warnings and added some that makes sense - such as using `pub(crate)` in private modules.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -82,7 +82,28 @@
 //!
 //! `os-arch` and `authors` also can't be overriden and are derived from the environment.
 
-#![warn(clippy::unneeded_field_pattern, clippy::match_bool, clippy::get_unwrap)]
+#![warn(
+    //clippy::cargo_common_metadata,
+    clippy::branches_sharing_code,
+    clippy::cast_lossless,
+    clippy::cognitive_complexity,
+    clippy::get_unwrap,
+    clippy::if_then_some_else_none,
+    clippy::inefficient_to_string,
+    clippy::match_bool,
+    clippy::missing_const_for_fn,
+    clippy::missing_panics_doc,
+    clippy::option_if_let_else,
+    clippy::redundant_closure,
+    clippy::redundant_else,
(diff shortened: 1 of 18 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
 /target
 **/*.rs.bk
 /.vscode
+commitlint.config.js
```
sassman, line 4: Could you please keep such personal preferences in your local gitignore?

src/lib.rs
```diff
@@ -82,7 +82,28 @@
 //!
 //! `os-arch` and `authors` also can't be overriden and are derived from the environment.
 
-#![warn(clippy::unneeded_field_pattern, clippy::match_bool, clippy::get_unwrap)]
+#![warn(
```
sassman, line 85: I wonder why we would need custom clippy configuration, and not use the reasonable defaults? 
Personal prefectures?

taurr: Well, as I see it - @MarcoIeni showed us that Clippy could actually spot quite a few cases that are considered good Rust practice - it just doesn't do it by default. Why it doesn't, I don't know.

Apart from the `.gitignore` file (which came about because I've been trying out commitlint in githooks to improve the commit messages), I actually think it's a shame these warnings are not enabled by default.

This PR is not exactly a must, but it does show that Clippy can help us even more, though it does require us to "opt in".
If nothing else, we should at least consider enabling some of them.

sassman: I think they are pretty neat. I tend to like them actually 🙂

taurr:
Just removed the change to `.gitignore` - learned that I could create a file `<repo>/.git/info/exclude` for a local only `.gitignore` version... cool, learning new stuff 😄

sassman:
Cool, but I was referring to a local but personal gitignore as you can read [here](https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer)

sassman (approved):

PR #444, opened 2021-09-02
Remove cargo dependency

messense:
cargo dependency was originally added in https://github.com/cargo-generate/cargo-generate/pull/119, but it doesn't seem to do what it was supposed to do now, so I think cargo-generate doesn't have to depend on it now.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,12 +3,6 @@
 version = 3
 
 [[package]]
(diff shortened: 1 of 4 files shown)


sassman (approved):
thanks @messense for the PR, looks great. 

Makes cargo-generate way more lighter than before.

PR #445, opened 2021-09-04
feat: Pre/Post hooks

### LINKED ISSUE #17: Custom variables.

### LINKED ISSUE #18: Pre/Post Hooks

### LINKED ISSUE #92: RFC: Ready-made configurations/presets for templates
In the embedded WG we have another challenge with templates to solve. Once we have #17, it would be nice to provide additional guidance for the user to break down sensible value combinations into presets which can be stored in the templates to vastly simplify the setup.

An example for a system like this exists in the C/C++ embedded world with [platform.io](https://platformio.org/) which (amongst other things) provides something very similar to `cargo-generate`. Instead of providing a lot of details (most will probably not even know from the top of their head) about the platform one is interested in, there're presets which can be selected, e.g.:

```
# platformio boards stm32

Platform: ststm32
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
ID                    MCU            Frequency  Flash   RAM    Name
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
1bitsy_stm32f415rgt   STM32F415RGT   168MHz    1MB     128KB  1Bitsy
b96b_f446ve           STM32F446VET6  168MHz    512KB   128KB  96Boards B96B-F446VE
armstrap_eagle1024    STM32F417VGT6  168MHz    1MB     192KB  Armstrap Eagle 1024
armstrap_eagle2048    STM32F427VIT6  168MHz    1.99MB  256KB  Armstrap Eagle 2048
...
```

Creating a new project using such a template will result in nicely configured environment which allows to jump right into development.

taurr:
closes #18
closes #92, as any complex logic for setting variables can be handled.
closes #17, for the same reason as #92, and as we already have full support for variables/placeholders. @japaric even created an issue in the embedded-wg waiting for this possibility.

This creates support for hooks (scripts) written in [`Rhai`] - the same engine that powers the conditionals feature.

Examples and docs are part of the commits :-)

[`Rhai`]: https://rhai.rs/book/about/index.html


src/ignore_me.rs
```diff
@@ -75,8 +75,12 @@ fn unwrap_path(it: Result<ignore::DirEntry, ignore::Error>) -> PathBuf {
     it.expect("Found invalid path: Aborting").into_path()
 }
 
-fn remove_dir_files(files: &[PathBuf], verbose: bool) {
-    for item in files.iter().filter(|file| file.exists()) {
+pub fn remove_dir_files<T: IntoIterator<Item = I>, I: Into<PathBuf>>(files: T, verbose: bool) {
```
sassman, line 78: suggestion: working with the `impl` syntax helps to improve the readability quite a lot.

so here we would end up having this:
```rust
pub fn remove_dir_files(files: impl IntoIterator<Item = impl Into<PathBuf>>, verbose: bool) {
```

But of course this is quite matter of taste, but this syntactical sugar was introduced for a reason right? :)

src/include_exclude.rs
```diff
@@ -7,15 +7,25 @@ use ignore::gitignore::{Gitignore, GitignoreBuilder};
 use std::path::Path;
 
 #[derive(Default)]
-pub struct Matcher(Option<MatcherKind>);
+pub struct Matcher(Option<MatcherKind>, Vec<String>);
+
+pub enum ShouldInclude {
```
sassman, line 12: praise: very nice usage of an enum

src/lib.rs
```diff
@@ -460,29 +463,39 @@ fn expand_template(
...
-            anyhow::bail!(ConversionError::MissingPlaceholderVariable {
-                var_name: slot.var_name.clone()
-            })
-        }
-        interactive::variable(slot, provided_value)
-    })?;
-    let template = add_missing_provided_values(template, template_values)?;
-
-    let (template_cfg, template) = merge_conditionals(template_config, template, args)?;
+    let liquid_object = template::create_liquid_object(name, &crate_type, args.force)?;
```
sassman, line 470: question: I would prefer to not have such random renaming of variables when there is no need for chnage, the names have all been choosen on purpose and I'd kindly ask you to respect the authors naming decisions and rather adapt your mental model. 

Ofc if you think that something is just pure wrong or nonsense then I'd suggest to keep those discussions into separated PRs, that would simplify the review and keep things pretty focused on the feature.

sassman (commented):
very great feature, thanks for all your hard work!
Pls find also some nits and a more of a general comment regarding renaming and scope of PR changes.

src/include_exclude.rs
taurr: thx :-)

src/lib.rs
taurr: If you refer to the `liquid_object` variable - I confess to having renamed it from the original name `template`.

Having read, and re-read, your comment a few times (because I am unsure what to make of it) - I think have a fair grasp of the issue at hand; it's about the fact I did a rename as part of this commit/PR, and not the actual rename - right?
**For that I am sorry**; I did not realize I should restrict each commit/PR that strictly - not used to that.

If on the other hand it's about the actual rename, here's my reasoning (and original reply prior to catching the alternate meaning/reason for the comment)...

**Reason for rename:** back in August, the `template::substiture(...)` function was renamed (by me, again!) to `create_liquid_object(...)` - at that time because *the function didn't do any substitution at all*, it just created and prepared a `Object` from the `liquid` crate - litterally.

What I did not notice at the time was the variable in question. It is not a template, nor does it represent one. It is a object for holding template values when `liquid` eventually does substitution in the actual template files - in the `walk_dir(...)` function.

Hence I renamed it to reflect this discovery.

**If I was wrong, or anyone fells hurt, I appologise here and now** - but I do feel the original name didn't fit the actual use anymore.

If you want me to rename it back, please just say so.


For the record: I myself *do not care if anyone renames or otherwise changes any part of any code I have ever written*. I actually expect the other contributors to do so, in order for them to better understand the code and/or for the code to be in the best shape it can be at any point in time.

sassman: > For that I am sorry; I did not realize I should restrict each commit/PR that strictly - not used to that.

no worries, no feelings hurt. All good. If renaming make sense, and your provided reasoning does make sense, then go for it. I'm not tied to names (only if I spend a fair amount of time to express a concept and this expression level would decrease because of the rename).

So sub notion was about focus on PRs, so this PR is already quite a commit 💣  which is great because of the nice feature, but the level of noise (coming from random function / variable renaming) might be (if possible) kept low.

So having that said, I totally can support this particular rename.

sassman (approved):
what a PR, 🥇

sassman:
@taurr thanks again for the great work. Could I ask you to write a little summary in the `CHANGELOG.md` that would fit best to the hooks feature? I feel not 100% confident to come up with something that would honor this feature and your contribution enough.

taurr:
My own comment in CHANGELOG.md is probably not good either - but its there.

I suspect this is one of those features that we may end up adding quite some functionality to over time.

sassman:
Yeah I think so too. This feature is a so mighty enabler, that it almost screens for a blog post to illustrate all the possible use cases that are hidden in this gem.

sassman (approved):

taurr:
> Yeah I think so too. This feature is a so mighty enabler, that it almost screams for a blog post to illustrate all the possible use cases that are hidden in this gem.

That would probably also be cause for a [TWIR](https://this-week-in-rust.org) entry.

Anyhow, I'm a lousy author, so *dips on not writing the blog post* 😆

PR #446, opened 2021-09-05
feat: prompt for subfolder upon multiple templates

taurr:
This lets the user specify the base path (or git repo) for a potential multi-template repo. If there's more than 1 template in the repo, the user is asked to select the one to expand.

Given the user specifies a template folder without `cargo-generate.toml`, look for the config in child dirs. Then depending upon the number of configs found do:

0 => use the user specified folder
1 => use the folder with the found config
2+ => prompt the user for which template to use.

t-rec_1

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -4,6 +4,7 @@ use serde::Deserialize;
 use std::path::Path;
 use std::{collections::HashMap, fs};
 use std::{convert::TryFrom, io::ErrorKind};
+use walkdir::WalkDir;
 
 pub const CONFIG_FILE_NAME: &str = "cargo-generate.toml";
 
@@ -60,6 +61,27 @@ impl Config {
             None => Ok(None),
         }
     }
+}
+
+pub fn locate_template_configs(dir: &Path) -> Result<Vec<String>> {
+    let mut result = vec![];
+
+    for entry in WalkDir::new(dir) {
+        let entry = entry?;
+        if entry.file_name() == CONFIG_FILE_NAME {
+            let path = entry
+                .path()
+                .parent()
+                .unwrap()
+                .strip_prefix(dir)
+                .unwrap()
+                .to_string_lossy()
+                .to_string();
+            result.push(path)
+        }
+    }
+
+    Ok(result)
 }
 
 #[cfg(test)]
--- a/src/config.rs
+++ b/src/config.rs
@@ -86,12 +86,46 @@ pub fn locate_template_configs(dir: &Path) -> Result<Vec<String>> {
 
 #[cfg(test)]
 mod tests {
+    use crate::tests::create_file;
+
     use super::*;
     use std::fs::File;
     use std::io::Write;
     use std::str::FromStr;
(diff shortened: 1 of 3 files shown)


taurr:
t-rec

sassman:
this is a very neat feature! I like it a lot. Could you maybe just add a little test case to cover this additional behaviour?

taurr:
> this is a very neat feature! I like it a lot. Could you maybe just add a little test case to cover this additional behaviour?

I'd love to - but to be honest I don't know how right now!

Are there any tests for the interactive part of `cargo-generate` at all?

How do we create tests for `cargo-generate` asking for values? Any ideas??

src/config.rs
```diff
@@ -62,6 +63,27 @@ impl Config {
     }
 }
 
+pub fn locate_template_configs(dir: &Path) -> Result<Vec<String>> {
```
sassman, line 66: a sort of a unit test could be added for this function

src/lib.rs
```diff
@@ -299,16 +302,32 @@ fn resolve_template_dir(template_base_dir: &TempDir, args: &Args) -> Result<Path
...
-                style("...").bold()
-            );
-            Ok(template_dir)
+            Ok(auto_locate_template_dir(&template_dir)?)
+        }
+        None => auto_locate_template_dir(template_base_dir.path()),
+    }
+}
+
+fn auto_locate_template_dir(template_base_dir: &Path) -> Result<PathBuf> {
```
sassman, line 311: and here for this function we could also add a unit test

sassman (commented):
So I'd suggest to work with unit (or integration tests, because the file system dependency is hard to mock) and not with a regular e2e test as we have them mostly in the codebase

src/config.rs
```diff
@@ -62,15 +63,70 @@ impl Config {
...
+    fn locate_configs_can_locate_paths_with_cargo_generate() -> anyhow::Result<()> {
+        let tmp = tempdir().unwrap();
+        create_file(&tmp, "dir1/Cargo.toml", "")?;
+        create_file(&tmp, "dir2/dir2_1/Cargo.toml", "")?;
+        create_file(&tmp, "dir2/dir2_2/cargo-generate.toml", "")?;
+        create_file(&tmp, "dir3/Cargo.toml", "")?;
+        create_file(&tmp, "dir4/cargo-generate.toml", "")?;
+
+        let separator = std::path::MAIN_SEPARATOR.to_string();
+        let expected = vec!["dir2/dir2_2".replace("/", &separator), "dir4".to_string()];
```
sassman, line 120: suggestion: instead of manually replace the path separator, you could simply use `Path::join()` that takes care of using the correct platform specific path separator. something like maybe

```rust
        let expected = vec![Path::new("dir2").join("dir2_2").to_string(), "dir4".to_string()];
```

taurr: As `Path` and `PathBuf` don't have a `to_string()` method, it required a little tinkering.

But done!

sassman (approved):

Issue #461, opened 2021-09-23
Error: ⛔   Git Error: invalid version 0 on git_proxy_options; class=Invalid (3)


B4TT3RY:
Not working after updating to 0.10.0 version.

```
cargo 1.55.0 (32da73ab1 2021-08-23)
cargo-generate 0.10.0
git version 2.33.0
```

taurr:
Duplicate of #456

Same reason... it's due to a dependency inside `git2`

sassman:
Patch will be released today.

sassman:
This should now be solved with v0.10.1

TonalidadeHidrica:
@taurr Thanks for the reply!  I'm using `macOS Big Sur 11.6`.  FWIW, `git 2.33.0`, which I currently use, is one installed via Homebrew, located at `/usr/local/bin/git`, while there is another OS-built-in git at `/usr/bin/git` of version `git version 2.30.1 (Apple Git-130)`.   `libgit2` has also been installed via `brew`, whose version is `libgit2: stable 1.3.0`.  I'm not sure there's a system-wide installation of libgit2.

taurr:
@TonalidadeHidrica thx, as I tried to explain, you version of `git` itself really doesn't matter - it is not used.
`cargo-generate` uses another Rust dependency called `git2` - which is the one using `libgit2` (at least I think it does - but I don't think it uses the system installed one as most Rust stuff is statically linked).

In other words, the problem lies somewhere in the `git2` rust crate or one of its dependencies. An issue has already been filed there.

TonalidadeHidrica:
Oh sorry, I didn't read it well.  So there's nothing we can do other than fixing `git2` crate?

taurr:
I'm not even sure it's a `git2` crate specific issue - seems to be a fairly wellknown issue: https://www.google.com/search?q=invalid%20version%200%20on%20git_proxy_options%3B%20class%3DInvalid%20(3)

TonalidadeHidrica:
I've not fully understand what you explain, but at least this is the result for `ldd`.  What other information do I need to supply?

```
/Users/user/.cargo/bin/cargo-generate:
        /usr/local/opt/libgit2/lib/libgit2.1.3.dylib (compatibility version 1.3.0, current version 1.3.0)
        /usr/local/opt/openssl@1.1/lib/libssl.1.1.dylib (compatibility version 1.1.0, current version 1.1.0)
[...]

Issue #455, opened 2021-09-16
`cargo test` shortcut for template authors


sassman:
As a **template author**
I want to **have an easy way to test my template code**
So that **I save time to manually generate a project, run then cargo test in it and delete the generated project again**

I imagine something like
```sh
cd my-wonderful-template
cargo generate --test

## will run cargo generate as a template user would do, but just into a temp folder
## maybe with default variables to not run interactively?
## will run cargo test in the final generated project
## will delete the temp folder
```

taurr:
@sassman: I've got a prototype for this one running now.

It runs `cargo test --all-features -- --nocapture` instead of creating the files in the current folder as it normally would - thus running the test inside a tmp folder that is deleted upon exit.

This works very nicely, except for one caveat; I can see myself wanting to specify different arguments to `cargo test`, such as `--bins` or similar. Any ideas as to how we could specify this?

example run from my windows VM:
```shell
> cargo run -- gen --git https://github.com/Ureakim/rust-template_basic --name myproject --test
    Blocking waiting for file lock on build directory
   Compiling cargo-generate v0.10.0 (C:\Users\johnn\Desktop\Projects\cargo-generate)
    Finished dev [unoptimized + debuginfo] target(s) in 7.33s
     Running `target\debug\cargo-generate.exe gen --git https://github.com/Ureakim/rust-template_basic --name myproject --test`
 Generating template ...
[1/4]   Done: Cargo.toml  
[2/4]   Done: rustfmt.toml
[3/4]   Done: src\main.rs 
[4/4]   Done: src
   Compiling myproject v0.1.0 (C:\Users\johnn\AppData\Local\Temp\.tmpQFeb3Y)
    Finished test [unoptimized + debuginfo] target(s) in 0.95s
     Running unittests (target\debug\deps\myproject-6548a729dd0db5ee.exe)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

> 
```

sassman:
that is also on interesting idea.
My proposal was more focussed on the idea, that you got a template in your `CWD` and you want to test this.

```sh
cargo generate test
```

or even with 
```sh
cargo generate test --all-features -- --nocapture
```

so that it would does also template the current template into a temp dir, runs just the subcommand as provided (`cargo test` or `cargo test --all-features -- --nocapture`) and removes the tmp dir afterwards.

taurr:
[...]

PR #462, opened 2021-09-23
revert_to_git2_0.13.21_to_avoid_issues

taurr:
This change is related to #461 and #455.
Both issues seem to be caused by the use of the newer `libgit2-sys` version.

An issue has been created for `git2` team: https://github.com/rust-lang/git2-rs/issues/746

It is a temporary solution that will cause `dependabot` to complain, but without this inconvenience `cargo-generate` outright fails to run for some users.


sassman (approved):
Fine for me

PR #467, opened 2021-09-29
improvement: use a choice prompt instead of strings for boolean placeholders

### LINKED ISSUE #450: improvement: use a choice prompt instead of strings for boolean placeholders
When specifying a placeholder of type bool, `cargo-generate` uses a prompt designed for strings.

To improve useability, consider using a choice prompt instead.

t-rec_1

taurr:
Fixes #450


taurr:
rebased!

taurr:
@sassman: dunno how to make Github re-evaluate this, but the error here is a DNS error, and NOT due to code.

sassman:
Yeah, that happens sometimes. I retriggered the build.

PR #472, opened 2021-09-29
fix: Rhai variable extension not updating Liquid Object

### LINKED ISSUE #471: Rhai variable extension doesn't work
I was testing cargo generate with Rhai scripting and reading the source code on this repo when I noticed that the variable extension should't work at all. In the source code `src/hooks/mod.rs` only an immutable reference to the Liquid object was handed to the code that interacts with rhai scripting. In the current implementation you should only be able to set and get Liquid variables inside the same Rhai script, but not set a variable that can be used in the Liquid template expansion. This also means I could't make the hooks example work!

I have already implemented a simple solution but I have never done an Open Source contribution and would like some comments on this topic. Maybe a have just missed something!

GabrielDertoni:
fixes #471

### DIFF
--- a/src/hooks/mod.rs
+++ b/src/hooks/mod.rs
@@ -27,7 +27,7 @@ impl<F: FnOnce()> Drop for CleanupJob<F> {
 
 pub fn execute_pre_hooks(
     dir: &Path,
-    liquid_object: &liquid::Object,
+    liquid_object: Rc<RefCell<liquid::Object>>,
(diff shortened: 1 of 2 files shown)


taurr (approved):
LGTM :-)

How on earth I overlooked this one when I implemented the feature is a mystery to me, but good catch.

sassman (approved):

sassman:
Thanks @GabrielDertoni for your contribution

PR #474, opened 2021-10-02
feature: add support for arrays in the variable::set Rhai script extension

### LINKED ISSUE #473: feature: in Rhai script, add support for array in variable::set
This feature is not an addition to the types of placeholders in general. Instead it aims only to allow the usage of arrays through Rhai scripts. Rhai already has support for arrays and so does Liquid, so the implementation is simple and minimal changes would be required.

Today there are only two variable types that can be sent from Rhai to Liquid: strings and booleans. With this feature we could also have an array of any of the types. This includes arrays of arrays and arrays with different types at each index. This still shouldn't be a problem since both Rhai and Liquid have support for arrays with dynamic types.

With this feature we could write a Rhai script that prompts the user for a comma separated list of dependencies which can then be put into Cargo.toml somewhere else in the template via Liquid.

```rhai
let dependencies = variable::prompt("Dependencies (comma separated list)")
    .split(",")
    .map(|input| {
        input.trim();
        input
    });

variable::set("dependencies", dependencies);
```

GabrielDertoni:
closes #473

### DIFF
--- a/src/hooks/variable_mod.rs
+++ b/src/hooks/variable_mod.rs
@@ -225,7 +225,7 @@ fn rhai_to_liquid_value(val: Dynamic) -> Result<Value> {
     val.as_bool()
         .map(Into::into)
         .map(Value::Scalar)
-        .or_else(|_| val.clone().as_string().map(Into::into).map(Value::Scalar))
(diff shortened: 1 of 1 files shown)


sassman:
@GabrielDertoni thanks for the fix, looks good now.

Issue #456, opened 2021-09-17
Unable to run tests on Manjaro after upgrading `git2` to 0.13.22


taurr:
Tracing the inability to run tests leads me to believe this is a `git2` issue.
I've created https://github.com/rust-lang/git2-rs/issues/746 on their end.

This is mostly for tracking and monitoring the issue.

taurr:
@sassman: I'm tempted to sugest we *degrade* `git2` back to the previous version.

Not only does the current version cause issues in the integration tests (I for one can't run them at all!), but according th #461 some users can't even run `cargo-generate` anymore.

sassman:
Patch is released. But let's keep this issue open as a tracking issue for the dependency that has the problem.

miku86:
Hey folks,

seems this is not completely solved on Arch.

When I use cargo for the installation (`cargo install cargo-generate`), the error still shows up.

```
cargo generate --git https://github.com/rustwasm/wasm-pack-template
⚠️   Unable to load config file: /home/user/.cargo/cargo-generate.toml
Error: ⛔   Git Error: invalid version 0 on git_proxy_options; class=Invalid (3)
```

When I delete `cargo-generate` from `$HOME/.cargo/bin` and install it with pacman, it works as expected.

```
Arch 5.10.68-1-lts
cargo 1.55.0 (32da73ab1 2021-08-23)
cargo-generate 0.10.1
git version 2.33.0
```

sassman:
seems we got more insights on #461 so will close this as duplicate of the other.

PR #477, opened 2021-10-09
fix(git+libgit2:versions): try to fix deps issue

sassman:
- as described in #461 the version pinning does not work as expected
- so we not pin the minor version at all and let the `cargo` at install time do the decisions (even they might be wrong), depending on the systems installed `libgit2` version


sassman:
I verified this patch on `archlinux:latest` and it works. So would fix #456 and #461

PR #488, opened 2021-11-07
fix: print regex error

MarcoIeni:
Before:
image

After:
image


sassman:
nice, could you pls share the `username` that you have used to provoke this error? I'd like to reproduce it very identically.

MarcoIeni:
yes, sure:

```toml
[placeholders.username]
type = "string"
prompt = "GitHub username (or organization)?"
regex = "/^[a-z\\d](?:[a-z\\d]|-(?=[a-z\\d])){0,38}$/i"
```

MarcoIeni:
Probably I need to rename `placeholders.username` to `placeholders.gh-username`, because I see that `{{username}}` is now reserved :)

PR #489, opened 2021-11-07
maintenance: whitelist `time` advisories

MarcoIeni:
This will fix the `audit` step that is failing in CI.

The same file is used in [tokio](https://github.com/tokio-rs/tokio/blob/master/.cargo/audit.toml)


sassman (approved):
Thanks, if it works for tokio then it should do it for us too.

PR #490, opened 2021-11-07
README: fix example regex

MarcoIeni:
In this way strings that start or end with a letter are not matched.
Otherwise strings like `a22` or `321b` are recognized as `phone_number`.


sassman (approved):
Thanks for taking care of that

Issue #522, opened 2022-01-16
support providing `project-name` via hooks, env vars and template value file


xoac:
Hi, 
I wanted to create a template that allows me to create a project within URL. The question about `project-name` in most cases is in GitHub link and this is in my opinion, the best approach in creating a new project from template. The `project-name` variable must be required, but checking about its absence should be postponed until all pre-hooks are called.

This is a pseudo dialogue I want to achieve:

**Past link to repository of your project:** <github.com/organizaton/my-awesome-rust-project>
**Is the name of your project correct `my-awesome-rust-project`?** [Y (Yes)/e (Edit)] 
**Your project is created!!**

To make this backward compatible the template should set options that allow it to set `project-name` on it's own. For example in `cargo-generate.toml` `template.project-name.ask = false` (default `true`)

What do you think about this idea?

sassman:
Interesting feature. 

Let me explain 2 arguments why I think that would be a bit counter productive.

1. all my templates have the name "template" in the url, and I don't want to be have the project to be called template.
2. `cargo-generate` supports to have multiple sorts of templates in one single git repository. For example you can have a folder `cli` and a folder `lib` in the repository. Then when using `cargo-generate` with this one repository url it depends on the a different command line option what folder will be taken. So therefore the repository URL would also not reflect the right project name.

xoac:
I think you misunderstood me, I suggest:
1. The cargo-generate require `{project-name}` to be defined (but not exactly as user input).
2. template is allowed to disable prompt for `{project-name}`
3. cargo-generate will exit with error if template will not provide `{project-name}` variable with pre-hook scripts.

Maybe examples would be more clear:
1. You want to create `randomly-named-rust-project-template`
2. cargo-generate ask about `{project-name}` (a bit ironic...)
3. You ignore user input and overwrite it with pre-hook script with random name for project

In my case I want to create template like this:
1. You run `cargo generate --git me/my-template`
2. You are asked about git origin of project  (so `github.com/me/my-awesome-project`)
3. `my-template` sets `{project-name}` to `my-awesome-project`

sassman:
Yeah, I have misunderstood you in the first place.

Could you please show me such a template that would provide the `project-name` via hook?

In general it make sense that the `project-name` is also possible to provide via
- environment variable: `CARGO_GENERATE_VALUE_PROJECT_NAME`
- a [template value file](https://github.com/cargo-generate/cargo-generate#---template_values_file-flag)
[...]

Issue #67, opened 2018-07-27
remove libgit2 in favor of shelling out to git


anirudhb:
This would remove the dependency on Git, and we would just link in libgit2. It would require less things to be installed, especially when you start distributing this as a binary, since Rust statically links everything!

killercup:
I think we are already using the git2 crate for everything in the main
binary, and it is a binding to libgit2

anirudhb:
Then I think your README is wrong :) See the dependency on `git`?

killercup:
It was added in #45. I'll look into it. For running the tests, the git CLI is required.

killercup:
@migerh, can you help me out here? In #45 you added that git is a requirement to building this crate. Was this because you wanted to run the tests, too, or does some other dependency require the git binary to exist?

If it's only because of the tests, I think this issue can be closed. Would you agree, @anirudhb?

migerh:
I was just collecting everything needed to compile cargo-generate and did not run any tests. But I also did not aim for a minimal list of dependencies. The minimal dependencies to build `libssh2-sys` on a vanilla `ubuntu:latest` docker  image are `cmake` and `zlib1g-dev`. Instead of describing how to install `zlib1g-dev` for macOs, Windows and various Linux distros I opted to link to the git install manual, because git apparently brings this already along (at least for ubuntu).

So no, the git binary is not required to build this crate.

migerh:
I only just now realized that I read lib**ssh2**-sys as lib**git2**-sys all this time 🤦‍♂️ Installing git just to get libssh2-sys to compile does not really make that much sense.

So I dug a little deeper. Turns out that the [git2](https://crates.io/crates/git2) crate brings libgit2 along and there's no need to install it on your system for git2 to compile:

> The source for libgit2 is included in the libgit2-sys crate so there's no need to pre-install the libgit2 library, the libgit2-sys crate will figure that and/or build that for you.

libgit2 apparently uses cmake to compile...

This raises the question why libssh2-sys can't just do the same? If that crate could bring zlib1g-dev along with it there were no need to mention it in the readme. Btw, there's also a [libz crate](https://crates.io/crates/libz-sys).

Freaky:
For what it's worth, I'd much rather a tool just call out to git if it's only doing simple stuff.

[...]

PR #547, opened 2022-02-13
Refactor of handling favorites, git and path

xoac:
Hi,

I wanted to implement #522 but I found it difficult. When I dig into code I found  that some of it should in my opinion be refactored:

The first problem is with `--git` flag. If you set it to local path it will act differently than `git clone` what is a little bit unexpected. This is because `src/git.rs` try to check if the value in `--git` is not a local path and if it's it will just copy it. `git2` already handle local path so there is no necessary of additional code. 

The second problem is with **favorite arg**. If there is no **favorite configuration** in config file. cargo-generate will use **favorite arg** as the `--git` flag. In my opinion we should here distinguish if this is a **git** or **path** and handle it differently.

I would also expect that:
```
cargo generate --git <repository> <template_name>
```
and
```
git clone <repository> <local_path>
cargo generate --path <local_path> <template_name>
```
work the same. But now when ` --path` is used it just ignore **subfolder arg** (and ask about witch template is should actually use)

https://github.com/cargo-generate/cargo-generate/blob/d95863bd5f05471c0c529a129f584af0b73f9857/src/lib.rs#L263-L272 
 
 I would suggest this approach:
 1. [ ] `--git` is handled as much as possible by `git2` library 
 2. [ ] `--path` is used when we want to plain copy of some location. 
 3. [ ] `favorite` should be decoded in this order
     1. look for defined favorites in configuration and use one if name match
     2. if there is a `--branch` flag used consider **favorite arg** as `--git`
     3. if **favorite arg** starts with abbreviation  consider it as `--git`
     4. try to use **favorite arg** as `--path` (check if such path exist, is dir etc.)
     5. fallback to `--git` 
 
This changes are against 0.12.0 ( I made them some day ago but decided to publish today) and at this moment are just an idea of refactoring if you find it right I will continue to work on it.
The current changes also contains additional level of credentials through ssh-agent.


sassman:
Hi @xoac,

let me try to clarify a couple of things.

> I would also expect that:
>
> cargo generate --git <repository> <template_name>
> 
> and
> 
> git clone <repository> <local_path>
> cargo generate --path <local_path> <template_name>
> 
> work the same. But now when --path is used it just ignore subfolder arg (and ask about witch template is should actually use)

`--path` works with a local path but very different than `git clone <local-path>` works and that is intended. For example `git clone` uses the `HEAD` that is committed. On the other side `cargo generate --path` takes the folder as is, that means if there are changes that are not commited in the template repo it will just take them. 

That makes it the perfect tool for locally testing an WIP template repo, without the need of committing the changes in the template repo over and over. 

> The second problem is with favorite arg. If there is no favorite configuration in config file. cargo-generate will use favorite arg as the --git flag. In my opinion we should here distinguish if this is a git or path and handle it differently.

that is also intended because it would allow to have the following resolution order for example `cargo generate sassman/demo`
- `sassman/demo` is first checked if there is a favorite config for it, if not
- it will check if there is any local folder matching it, if not
- it checks if there is GitHub repo that matches it

so far that has been working out quite nicely, at least for me and others.

But I do see that the behaviour might be confusing, especially when there is `--path` that should be explicitly used.

xoac:
> --path works with a local path but very different than git clone <local-path> works and that is intended. For example git clone uses the HEAD that is committed. On the other side cargo generate --path takes the folder as is, that means if there are changes that are not commited in the template repo it will just take them.

I don't want anything different. I want very similar behavior (to current one) - mostly refactor just the code. 

> that is also intended because it would allow to have the following resolution order for example cargo generate sassman/demo

I want parse `<favorite>` to try guess as correct as possible (current behavior) what user want but assume that user know what is doing when it pass `--git` or `--path` and not modify his input. 

I will try my best to explain myself but I still learning English. 

## Example of current usage (as of 0.12.0)

Some simple scenarios that should help understand what is in my mind. I have cloned https://github.com/xoac/cargo-generate-two-templates into `xoac/cargo-generate-two-templates`. (The output below contains 3 levels of structure tree). 
```
$ tree -L 3
.
└── xoac
    └── cargo-generate-two-templates
        ├── logo.svg
        ├── template
        └── template2
```
_This is actually how I store all project on my computer `~/projects/github.com/<org-name>/<repo>`_

Try to predict the output. Be aware that `xoac/cargo-generate-two-templates` is a valid local path and valid org/repo on github:
1. `cargo generate --git xoac/cargo-generate-two-templates template` should clone the github repo, the local repo, or plain path (what you expect)?
2. `cargo generate --git github.com/xoac/cargo-generate-two-templates` should it work?
3. `cargo generate --path xoac/cargo-generate-two-templates template`
    <details>
    
    image
    I have already specified subfolder in CLI
    </details>
4. `cargo generate --git xoac/cargo-generate-two-templates template` the same out as with `--path` or different?
5. know better that user wanted path rather than git and this is why he wrote `--git`
    ```
    cp -r xoac/cargo-generate-two-templates/template local-template-without-git
    cargo generate --git local-template-without-git
    ```
    <details>
        
    The output could be. **The `<path>` exists but it's not a git repository. Maybe you wanted `--path`?**
    </details>
6. `cargo generate --path https://github.com/xoac/cargo-generate-two-templates `
    <details>
     
    why `--git` assume that user could wanted path but `--path` do not assume it can wanted git? This approach could nearly lead as to a place where `--git` `--path` and `<favorite>` could be all the same.
    </details>

Now the harder part `cargo generate  xoac/cargo-generate-two-templates` (so `<favorite>`) should expend to?

1.  `xoac/cargo-generate-two-templates` is first checked if there is a favorite config for it, if not
2. it will check if there is any local folder matching it, (if yes, will it be equivalent of **`cargo generate --git` or `cargo generate --path`**)?
3. it checks if there is GitHub repo that matches it 

So the problem is that `<favorite>` is not parsed instead is [directly mapped into `--git`](https://github.com/cargo-generate/cargo-generate/blob/v0.12.0/src/favorites.rs#L81)  

## Proposal

1. `cargo generate --git <repo>` should only accept what `git clone` would accept
2. `cargo generate --path <local_path>` should use local files, but sub-folder should also work as with `--git`
3. `cargo generate <favorite>` (no configuration) should be parsed:
    1. If contains abbreviation like `gh`, `gl` etc. and if create a correct link (use `cargo generate --git <link>`)
    2. Check if `<favorite>` could be a valid path to template on local system if yes then should behave exactly like `cargo generate --path <favorite>`
    3. use it as user would call `cargo generate --git <favorite>`

### Braking changes

This will make the idea simpler (and very close to the current behavior). The braking changes are:
1.  `--git` will be more restricted but user still can use cargo generate without this flag and get old behavior. 
2.  `org/repo` will not be parsed as github - this complicate parsing and promote Microsoft for no reason (this is optional change)

## Motivation
This crate was evolved in quite a big piece of good software now. I think it would be nice to improve a little quality of code. 

Here are the main modification that I would like to made. Just git module was the simplest for me to start:

1. The user input `args: Args` is [modified](https://github.com/cargo-generate/cargo-generate/blob/d95863bd5f05471c0c529a129f584af0b73f9857/src/lib.rs#L79) and passed to all other modules what make them less independent. (It's hard to jump into one module - there is a need to understand all code to work with one). So it will be harder and harder to adding new functionality.
2. The code that works with `--path` and `--git` is very similar and could be unified or simplified. For example [first time](https://github.com/cargo-generate/cargo-generate/blob/v0.12.0/src/lib.rs#L139) dealing with local path and the [second](https://github.com/cargo-generate/cargo-generate/blob/v0.12.0/src/git.rs#L50)
3. The modules should be independent as much as possible. Should be passed variables instead of `Args` (it is difficult to know which variables are used in function).
4. Make possible in less line of documentation explain how `<favorite>` , `--git` and `--path` work with each other

sassman:
Thanks again for the very detailed and well written explanation. And I agree to almost everything you've mentioned.
The one exception we should really keep as is at the moment:

> 2. org/repo will not be parsed as github - this complicate parsing and promote Microsoft for no reason (this is optional change)

I argue to keep this behavior with mostly convenience and the fact that most users use template from github and also the most easiest entry point is to not to tell cargo-generate to use `--git` because a short form of `org/repo` means already implicitly it should be used git. 

All the other arguments 
- to make `--path` more strict
- make `--path` and `--git` accept both the subfolder inside the repo
- etc..

Make sense to me. 

When it comes to refactoring, I appreciate the desire to reduce dependencies and improve the code structure.
Some modules did really grow over time and have not the best decoupling. 

However please keep in mind that also other PRs are in the making and a very huge refactoring will put quite some burden to the people that have to review it, and the others that have to deal with the merge of the refactoring. 
This is why I would propose a minimal invasive approach, for example focus for now only on the `git.rs` module, introduce new structs and not rewrite the whole thing.

So finally I'm very curious how this will work out. In the end refactorings should be always very reasonable and the tests should a) not break and b) must be written for new paths and code.

xoac:
So some updates:
1. `github.com/org/repo` is not accepted with git clone - it was my bad. But we could implemented `cargo generate github.com/org/repo` to be handled as `cargo generate --git git://github.com/org/repo.git` in seperate PR if you think this should be correct behavior.
2. Still WIP but wanted to show what is direction to get feedback about question I asked in code. Not yet necessary to review everything. 
3. `--git`, `--path` and `<favorite>` are handled as we agreed.

The general changes:
1. It was impossible to deal only with git module. I refactored the `favorite`, `path`, and `git`. I have created a module `user_parsed_input` to distinguish user input (cli, env, config files) from already parsed. Args is no longer mut.
2. **All old test are passing.** I added additional one that pass with `CARGO_GENERATE_GIT_WITH_CLI=true cargo test`
3. Some returning error changed.

> However please keep in mind that also other PRs are in the making and a very huge refactoring will put quite some burden to the people that have to review it, and the others that have to deal with the merge of the refactoring.

Yep my first proposal was against 0.12 so I know the pain. But I think this changes will make maintaining code easier

src/git/utils.rs
```diff
@@ -119,401 +23,117 @@ pub fn canonicalize_path(p: impl AsRef<Path>) -> Result<PathBuf> {
...
+) -> anyhow::Result<(TempDir, String)> {
+    let temp_dir = tempfile::tempdir()?;
 
-pub fn init(project_dir: &Path, branch: &str, force: bool) -> Result<Repository> {
-    fn just_init(project_dir: &Path, branch: &str) -> Result<Repository> {
-        let mut opts = RepositoryInitOptions::new();
-        opts.bare(false);
-        opts.initial_head(branch);
-        Repository::init_opts(project_dir, &opts).context("Couldn't init new repository")
+    let mut cmd = Command::new("git");
```
sassman, line 63: tbh it would be great if we would not shell out to git. I put already some arguments in the comment on the issue
https://github.com/cargo-generate/cargo-generate/issues/67#issuecomment-1032407568

sassman:
So the general refactoring looks good to me. 
However, I must say that I very much dislike the idea to shell out to git. 

IHMO: a tool like `cargo-generate` should not depend on ANY other binary that a user should have installed. That's the great thing about libraries that we can use to embed functionality and reuse great things.

xoac:
I agree but

1. There will probably be a long time before **full**, **working** git implementation will be provided in rust
2. For example now cargo-generate will **NOT** honorary:
   1. ssh-agent authorization
   2. most of .gitconfig configuration
   3. will no ask about passphrase to encrypted ssh-keys that is specified with `--identity`
   4. probably more less common situations
3. Most of users use git. (The same argument as most users use github.com - for `org/repo` worked ;) )

----
On the other hand

1. By default you still use git2 implementation
2. Using `CARGO_GENERATE_GIT_WITH_CLI=true cargo generate` will just work as expected by git user. No need to create issues and struggle with git2 details. 
3. Document `CARGO_GENERATE_GIT_WITH_CLI=true` as solution that will be removed in the future when full git implementation will be provided internally. Or even document it as experimental
4. Testing `CARGO_GENERATE_GIT_WITH_CLI=true cargo test` -- and that's all

tests/integration/config_file/defaults.rs
```diff
@@ -22,7 +22,7 @@ fn it_uses_ssh_identity_from_defaults_config() {
                 id = fake_ssh_id.display().to_string().escape_default()
             ),
         )
-        .file("fake_ssh_identity", "random foo")
+        .file("fake_ssh", "random foo") // FIXME: what is for?
```
xoac, line 25: I do not know what this is for? Please make sure I didn't break this test.

src/lib.rs
```diff
@@ -151,58 +156,42 @@ pub fn generate(mut args: Args) -> Result<()> {
...
+
+            let (temp_dir2, branch2) = clone_fn(git.url(), git.branch(), git.identity())?;
+            temp_dir = temp_dir2;
+            branch = branch2;
         }
-    }
-    Ok(())
+        TemplateLocation::Path(path) => {
+            temp_dir = copy_path_template_into_temp(path)?;
+            branch = String::from(DEFAULT_BRANCH); // FIXME is here any reason to set branch when path is used?
```
xoac, line 190: When `--path` is used we copy template in tmp and returning `branch`. But there is no branch actually

src/lib.rs
```diff
@@ -220,49 +209,47 @@ fn resolve_project_name(args: &Args) -> Result<ProjectName> {
...
+            fs::canonicalize(template_base_dir.join(subfolder)).with_context(|| {
+                format!(
+                    "not able to find subfolder '{}' in source template",
+                    subfolder
+                )
+            })?;
+
+        //TODO: Add comments to understand this check
+        // is this preventing when subfolder contains for example `../../name`?
+        if !template_dir.starts_with(&template_base_dir) {
```
xoac, line 225: Do I understand this correctly? That it prevent from `subfolder arg` being `../../../some-name`?

sassman: yes, that is a sandbox check mechanism, so that the repository cannot be escaped

src/lib.rs
sassman: the normal git clone mechanism yields the branch, so this must also provide a branch

xoac:
Let me know what is your decision about `CARGO_GENERATE_GIT_WITH_CLI` if you want it included then some documentation should be also added

src/lib.rs
xoac: Oh. I mean this should probably return `Option<String>` instead of `String`. If the branch is returned it's surprise that there is actually no git repository. 

And it would be more clear what's going on [here]( https://github.com/xoac/cargo-generate/blob/7e19fd8673c43fb24b19e96e876b0b621c427644/src/lib.rs#L146):
```rust
    if !args.init || args.force_git_init {
        info!("{}", style("Initializing a fresh Git repository").bold());
        args.vcs
            .initialize(&project_dir, branch, args.force_git_init)?;
    }
```
and replace `branch` with `branch.unwrap_or(DEFAULT_BRANCH)` 

----

For me the whole behavior here is surprise. I would expect that if I call `cargo gen --branch web3 example-org/example-template` it will init `main` branch instead of `web3`. What you think about `--init--branch` flag with default value `main` for initialize git repository? (I mean in other PR).

sassman:
> Let me know what is your decision about `CARGO_GENERATE_GIT_WITH_CLI` if you want it included then some documentation should be also added

Actually, I'd really appreciate if we could handle that as a feature in a followup PR.

That way the review would also be easier.

taurr:
Another issue, now that we're touching the CLI stuff... I know this is a breaking change, but our current use of clap/structopt is not utilizing subcommands properly, which stops us from defining options that are suitable for only some commands etc... it would be nice to define it properly :-/

xoac:
>Actually, I'd really appreciate if we could handle that as a feature in a followup PR. 

100% agree. I removed code for optional git from shell.

src/user_parsed_input.rs
```diff
@@ -0,0 +1,284 @@
...
+        }
+
+        // there is no specified favorite in configuration
+        // this part try to guess what user wanted in order:
+        // 1. look for abbrevations like gh:, gl: etc.
+        // 2. check if template directory exist
+        // 3. check if the input is in form org/repo<> (map to github)
+        // 4. assume user wanted use --git
+
+        let temp_location = if let Some(git_url) = abbreviated_git_url_to_full_remote(fav_name) {
```
sassman, line 115: could you please look at the clippy lint:

```
error: use Option::map_or_else instead of an if let/else
   --> src/user_parsed_input.rs:115:29
    |
115 |           let temp_location = if let Some(git_url) = abbreviated_git_url_to_full_remote(fav_name) {
    |  _____________________________^
116 | |             TemplateLocation::from(GitUserInput::with_git_url_and_args(&git_url, args))
117 | |         } else if let Some(path) = local_path(fav_name) {
118 | |             TemplateLocation::Path(path)
...   |
127 | |             ))
128 | |         };
    | |_________^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#option_if_let_else
help: try
    |
115 ~         let temp_location = abbreviated_git_url_to_full_remote(fav_name).map_or_else(|| if let Some(path) = local_path(fav_name) {
116 +             TemplateLocation::Path(path)
117 +         } else if let Some(git_url) = abbreviated_github(fav_name) {
118 +             TemplateLocation::from(GitUserInput::with_git_url_and_args(&git_url, args))
119 +         } else {
120 +             TemplateLocation::from(GitUserInput::new(
  ...
```

sassman (changes requested):
during the review I found some things that would require a bit of further work

src/user_parsed_input.rs
xoac: This is a little bit complicated. This is what clippy suggested:

```rust
let temp_location = abbreviated_git_url_to_full_remote(fav_name).map_or_else(
    || {
        local_path(fav_name).map_or_else(
            || {
                abbreviated_github(fav_name).map_or_else(
                    || {
                        TemplateLocation::from(GitUserInput::new(
                            fav_name,
                            args.branch.clone(),
                            ssh_identity,
                            args.force_git_init,
                        ))
                    },
                    |git_url| {
                        TemplateLocation::from(GitUserInput::with_git_url_and_args(
                            &git_url, args,
                        ))
                    },
                )
            },
            TemplateLocation::Path,
        )
    },
    |git_url| TemplateLocation::from(GitUserInput::with_git_url_and_args(&git_url, args)),
)
```
I will refactor it to make clippy happy but still not make this so nested.

src/git/mod.rs
```diff
@@ -1,5 +1,167 @@
...
+
+/// remove context of repository by removing `.git` from filesystem
+pub fn remove_history(project_dir: &Path) -> io::Result<()> {
+    let git_dir = project_dir.join(".git");
+    if git_dir.exists() && git_dir.is_dir() {
+        let mut attempt = 0_u8;
+
+        loop {
+            attempt += 1;
+            if let Err(e) = fs::remove_dir_all(&git_dir) {
```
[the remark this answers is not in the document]
xoac: I changed this unintentionally. Sorry for this.

xoac:
Rebase seems hard at this moment to resolve conflicts. Can I just merge main?

sassman:
> Rebase seems hard at this moment to resolve conflicts. Can I just merge main?

When this PR lands I will create a squash commit anyways.
So in fact a merge commit is not a problem.

sassman:
@xoac Thank you very much for this PR.
I created a new feature branch `feat/favourite-git-path-refactoring` where we can further fine tune certain pieces.
My intention is to have it merged very soon to `main`

PR #563, opened 2022-03-21
Bump git2 version to 0.14

TonalidadeHidrica:
Due to rust-lang/git2-rs#824, specifying `--branch` causes an error, making unable to use.  It seems this can be fixed by bumping the version to 0.14, so I bumped the version.


sassman (approved):

sassman:
@TonalidadeHidrica thanks for your contribution!

Issue #384, opened 2021-07-26
change ssh private key default to ed25519 before checking for rsa


jashandeep-sohi:
Currently, only RSA keys are supported: https://github.com/cargo-generate/cargo-generate/blob/528bf381807201bf91103dd3ca50bcca90365a5c/src/git.rs#L119

However, more and more people are defaulting to `ed25519`(myself included). Having to setup a `rsa` key just for `cargo-generate` is not ideal.

`cargo-generate` should at least support `ed25519` keys as well.

Edit: I realize I could manually override this via `-i | --identity`, but my request is to make it automatically use `id_ed25519` if it exists.

sassman:
My proposal would be to try a couple of defaults and what ever is there will be used. With this given order:
- RSA id_rsa
- ed25519

Anything else that folks are using that is not mentioned in the list above?

jashandeep-sohi:
This also got me thinking about something else. What happens when keys are password protected (or more recently require a security key to unlock). Wouldn't it just be better to delegate all of this to `ssh-agent` via https://docs.rs/git2/0.7.5/git2/struct.Cred.html#method.ssh_key_from_agent or any credential helper via https://docs.rs/git2/0.7.5/git2/struct.Cred.html#method.credential_helper

sassman:
those are very valid points. I'll toy around with them in order to understand the behaviour in cases there no `ssh-agent` is running or the credential helper fails to figure things out. 

But I guess we can extend the scope of this issue to support also arbitrary protected private keys, not only elliptic curve keys.

sassman:
with #1265 the default mechanism on unix/macOS is to use the function `add_default_ssh_keys` that adds a couple of default keys to look up first.

https://github.com/de-vri-es/auth-git2-rs/blob/fd6502e20b9e82063b950ac9e32f6454341e71f2/src/lib.rs#L344

PR #595, opened 2022-05-12
Enhance canonicalize_path and git errors

andrewjstone:
`canonicalize_path` will return an error if the constructed path does not exist,
because this is what `Path::canonicalize` does. However, in many places the use of `?`
short circuits the error return preventing the sometimes redundant
`path.exists()` check from running on the next line and bailing with the actual
path. This small change ensures that the path always gets printed when it does not exist.

Additionally, since this error occurs when `$HOME/.ssh/id_rsa` is not found, the user
is now made aware that the program is looking for this file. To help rectify this a help
message is appended in this situation.

Before:
```
cargo-generate git:(main) cargo generate --git git@github.com:xxx/xxx --name yyy
⚠️   Unable to load config file: /Users/ajs/.cargo/cargo-generate.toml
Error: ⛔   Git Error: path does not exist
```

After:
```
cargo-generate git:(main) cargo generate --git git@github.com:xxx/xxx --name yyy
⚠️   Unable to load config file: /Users/ajs/.cargo/cargo-generate.toml
Error: ⛔   Git Error: path does not exist: /Users/ajs/.ssh/id_rsa
help: try using the '--identity' flag
```


andrewjstone:
This is not really a fix for https://github.com/cargo-generate/cargo-generate/issues/384, but it helps reduce confusion to new users by giving a more informative error message.

I'm happy to iterate on a better solution.

sassman (approved):
LGTM 🚀

sassman (commented):
fix merge conflict

PR #597, opened 2022-05-19
Less noise

v-n-suadicani-issuu:
Simply removes the "Unable to load config file" warning. I personally find this quite noisy and unnecessary.

### DIFF
--- a/src/app_config.rs
+++ b/src/app_config.rs
@@ -38,21 +38,17 @@ impl TryFrom<&Path> for AppConfig {
     type Error = anyhow::Error;
 
     fn try_from(path: &Path) -> Result<Self, Self::Error> {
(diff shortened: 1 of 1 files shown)


sassman:
could you pls check if these changes here are still required when PR https://github.com/cargo-generate/cargo-generate/pull/572 lands?

v-n-suadicani-issuu:
> could you pls check if these changes here are still required when PR #572 lands?

I tried checking out that PR and ran a test template and I still get:

```
⚠️   Unable to load config file: /home/victor/.cargo/cargo-generate.toml
🔧   Generating template ...
[...]
```

So no, it still appears.

sassman (approved):

Issue #596, opened 2022-05-16
Support for system commands


v-n-suadicani-issuu:
This is something provided by other templating libraries, such as cookiecutter. There are plenty of times it might be useful to run system scripts for a template.

It could be provided by exposing the [`Command`](https://doc.rust-lang.org/std/process/struct.Command.html) struct in the standard library to the Rhai scripts somehow. That would allow you to use the placeholder variables and all to decide whether to run some system command.

This is obviously a security risk too, but presumably people are creating a template because they are going to run `cargo` on it - in which case a malicious `build.rs` could do whatever it wants anyway. So I don't think this really increases the security risk much, if at all.

taurr:
Except that `cargo` is not run on the template by `cargo-generate`, so the developer has the possibility (and some would say responsibility) to check the code before actually trying to build it.

I do agree that this would be nice, but I would have a dialog confirming whether it is OK to execute the commands before actually doing it - as well as a commandline option to resolve the dialog beforehand (so the expansion doesn't ask but just executes/denied as requested)

sassman:
> I do agree that this would be nice, but I would have a dialog confirming whether it is OK to execute the commands before actually doing it

I agree. At least for me as a user running arbitrary code on my machine provided by templates from the internet causes quite a bit of discomfort.

But of course, everyone can vote on this in order to express the usefulness.

v-n-suadicani-issuu:
>I would have a dialog confirming whether it is OK to execute the commands before actually doing it

I think that sounds perfectly reasonable - if integrated in the Rhai scripting, I suppose the command function could return an `Option` of [`Output`](https://doc.rust-lang.org/std/process/struct.Output.html), with it being set to `None` if the user refused to run the command. Then the script could handle that somehow if needed.

v-n-suadicani-issuu:
@taurr @sassman Took a stab at implementing this in #598, do let me know what you think.

sassman:
looks very good and it's now released in `v0.14.0` but I just realized that we miss proper docs for the feature as in the mdbook.

PR #598, opened 2022-05-19
Allow system commands

v-n-suadicani-issuu:
An attempt at #596.

I'm unsure if the command in Rhai should even try to return anything or if it should just "call and forget". But it might be useful to have the stdout (maybe even also stderr or whatever).

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -125,6 +125,12 @@ pub struct Args {
     /// Will enforce a fresh git init on the generated project
     #[structopt(long)]
     pub force_git_init: bool,
+
+    /// Allows running system commands without being prompted.
+    /// If this flag is not set, you will be prompted before running any
+    /// system commands in templates, as a security precaution.
+    #[structopt(short, long)]
+    pub allow_commands: bool,
 }
 
 #[derive(Debug, StructOpt, Clone, Copy)]
--- a/src/args.rs
+++ b/src/args.rs
@@ -127,8 +127,8 @@ pub struct Args {
     pub force_git_init: bool,
 
     /// Allows running system commands without being prompted.
-    /// If this flag is not set, you will be prompted before running any
-    /// system commands in templates, as a security precaution.
+    /// Warning: Setting this flag will enable the template to run arbitrary system commands without user confirmation.
+    /// Use at your own risk and be sure to review the template code beforehand.
     #[structopt(short, long)]
     pub allow_commands: bool,
 }
--- a/src/hooks/mod.rs
+++ b/src/hooks/mod.rs
@@ -5,10 +5,11 @@ use std::cell::RefCell;
 use std::rc::Rc;
 use std::{env, path::Path};
 
-use crate::config;
 use crate::emoji;
+use crate::{config, Args};
 
 mod file_mod;
+mod system_mod;
 mod variable_mod;
 
 struct CleanupJob<F: FnOnce()>(Option<F>);
@@ -26,20 +27,22 @@ impl<F: FnOnce()> Drop for CleanupJob<F> {
 }
 
 pub fn execute_pre_hooks(
+    args: &Args,
     dir: &Path,
     liquid_object: Rc<RefCell<liquid::Object>>,
     template_cfg: &mut config::Config,
 ) -> Result<()> {
-    let engine = create_rhai_engine(dir, liquid_object);
+    let engine = create_rhai_engine(args, dir, liquid_object);
     evaluate_scripts(dir, &template_cfg.get_pre_hooks(), engine)
 }
 
 pub fn execute_post_hooks(
+    args: &Args,
     dir: &Path,
     liquid_object: Rc<RefCell<liquid::Object>>,
     template_cfg: &config::Config,
 ) -> Result<()> {
-    let engine = create_rhai_engine(dir, liquid_object);
+    let engine = create_rhai_engine(args, dir, liquid_object);
     evaluate_scripts(dir, &template_cfg.get_post_hooks(), engine)
 }
 
@@ -65,7 +68,11 @@ fn evaluate_scripts(dir: &Path, scripts: &[String], engine: rhai::Engine) -> Res
     Ok(())
 }
 
-fn create_rhai_engine(dir: &Path, liquid_object: Rc<RefCell<liquid::Object>>) -> rhai::Engine {
+fn create_rhai_engine(
+    args: &Args,
+    dir: &Path,
+    liquid_object: Rc<RefCell<liquid::Object>>,
+) -> rhai::Engine {
     let mut engine = rhai::Engine::new();
 
     let module = variable_mod::create_module(liquid_object);
@@ -73,6 +80,9 @@ fn create_rhai_engine(dir: &Path, liquid_object: Rc<RefCell<liquid::Object>>) ->
 
     let module = file_mod::create_module(dir);
     engine.register_static_module("file", module.into());
+
+    let module = system_mod::create_module(args);
+    engine.register_static_module("system", module.into());
 
     engine.register_result_fn(
         "abort",
--- a/src/hooks/mod.rs
+++ b/src/hooks/mod.rs
@@ -5,8 +5,8 @@ use std::cell::RefCell;
 use std::rc::Rc;
 use std::{env, path::Path};
 
+use crate::config;
 use crate::emoji;
-use crate::{config, Args};
 
 mod file_mod;
 mod system_mod;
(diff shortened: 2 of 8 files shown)


src/hooks/system_mod.rs
```diff
@@ -0,0 +1,66 @@
...
+        // Only accept clearly positive affirmations.
+        matches!(
+            value.map(|s| s.trim().to_ascii_lowercase()).as_deref(),
+            Ok("yes" | "y")
+        )
+    };
+
+    if !should_run {
+        return Ok(Dynamic::UNIT);
+    }
```
v-n-suadicani-issuu, line 58: Also unsure about this - maybe it's better to just panic?

src/args.rs
```diff
@@ -125,6 +125,12 @@ pub struct Args {
     /// Will enforce a fresh git init on the generated project
     #[structopt(long)]
     pub force_git_init: bool,
+
+    /// Allows running system commands without being prompted.
+    /// If this flag is not set, you will be prompted before running any
```
taurr, line 130: I would prefer this doc to be written as what happes when the flag is set!

In other words; I'd like this more:
*Warning*: Setting this flag will enable `cargo-generate` to run system commands _without_ user confirmation. Use at your own risc.

src/hooks/mod.rs
```diff
@@ -26,20 +27,22 @@ impl<F: FnOnce()> Drop for CleanupJob<F> {
 }
 
 pub fn execute_pre_hooks(
+    args: &Args,
     dir: &Path,
     liquid_object: Rc<RefCell<liquid::Object>>,
     template_cfg: &mut config::Config,
 ) -> Result<()> {
-    let engine = create_rhai_engine(dir, liquid_object);
+    let engine = create_rhai_engine(args, dir, liquid_object);
```
taurr, line 35: You don't really need the entire `args` struct here - only the `allow_commands` flag.

src/hooks/mod.rs
```diff
@@ -26,20 +27,22 @@ impl<F: FnOnce()> Drop for CleanupJob<F> {
...
 }
 
 pub fn execute_post_hooks(
+    args: &Args,
     dir: &Path,
     liquid_object: Rc<RefCell<liquid::Object>>,
     template_cfg: &config::Config,
 ) -> Result<()> {
-    let engine = create_rhai_engine(dir, liquid_object);
+    let engine = create_rhai_engine(args, dir, liquid_object);
```
taurr, line 45: You don't really need the entire `args` struct here - only the `allow_commands` flag.

src/hooks/system_mod.rs
```diff
@@ -0,0 +1,66 @@
...
+};
+
+type Result<T> = anyhow::Result<T, Box<EvalAltResult>>;
+
+/// Creates the system module, containing the `command` function,
+/// which allows you to run system command.
+pub fn create_module(args: &Args) -> Module {
+    let mut module = Module::new();
+
+    let allow_commands = args.allow_commands;
```
taurr, line 17: See, only `allow_commands` is used 😃

src/hooks/system_mod.rs
```diff
@@ -0,0 +1,66 @@
...
+    module
+}
+
+fn run_command(name: &str, args: rhai::Array, allow_commands: bool) -> Result<Dynamic> {
+    let args: Vec<String> = args.into_iter().map(|arg| arg.to_string()).collect();
+
+    // If the user specified the --allow-commands flag, don't prompt.
+    let should_run = allow_commands || {
+        let prompt = format!(
+            "The template is requesting to run the following command. Do you agree?\n{} {}",
```
taurr, line 31: FYI: New style is allowed after rustc 1.60: 
````rust
let prompt = {
    let arguments = args.join(" ");
    format!("The template is requesting to run the following command. Do you agree?\n{name} {arguments}")
};
```

src/hooks/system_mod.rs
taurr: No, please return a proper error desribing the reason for the failure: User has denied execution of external command.

`cargo-generate` might be run in 'silent' mode, thus the prompt is disabled and the user never got to see it!

src/hooks/system_mod.rs
```diff
@@ -0,0 +1,66 @@
...
+
+    if !should_run {
+        return Ok(Dynamic::UNIT);
+    }
+
+    let output = Command::new(name).args(args).output();
+
+    Ok(match output {
+        Ok(o) => Dynamic::from_blob(o.stdout),
+        Err(_) => Dynamic::UNIT,
```
taurr, line 64: Give back `stderr` in case of err?

tests/integration/hooks.rs
```diff
@@ -20,6 +20,12 @@ fn it_runs_scripts() {
             file::rename("POST-TEST", "POST");
         "#},
         )
+        .file(
```
taurr, line 23: In general: more testing is needed.

At least check the behavior with and without `allow_commands`, as well as in the 'silent' mode (can't remember the flag). Should also test with failing commands.

taurr (changes requested):
Generally I like this, but I do have a few small issues that should be looked into - nothing big though.

src/hooks/mod.rs
v-n-suadicani-issuu: It felt a little strange to only propagate the `allow_commands` flags through, and I'm sure the compiler can optimize it anyway - but sure.

src/hooks/system_mod.rs
v-n-suadicani-issuu: There is no stderr in case of `Err` - the whole command failed to execute in that case.

src/hooks/system_mod.rs
v-n-suadicani-issuu: >return a proper error desribing the reason for the failure

I'm assuming you mean return from cargo-generate entirely? My thoughts were initially that the command could fail but only return to the Rhai script (the Rhai script could then handle the failing command however it wants) but I think this probably won't be very useful and you could always simulate it by running a bash script as the system command.

I added some handling for the silent mode :)

tests/integration/hooks.rs
v-n-suadicani-issuu: I added some tests, including one for silent mode and for a failing command :)

However, I'm not sure how to test the prompt - is there any way to do this? I've tried spawning the command as a subprocess and writing stuff to stdin, but it doesn't seem to work. I'm not sure if such a test would be strictly necessary but it would be nice.

src/hooks/mod.rs
```diff
@@ -9,12 +9,13 @@ use crate::config;
...
 
 mod file_mod;
+mod system_mod;
 mod variable_mod;
 
 struct CleanupJob<F: FnOnce()>(Option<F>);
 
 impl<F: FnOnce()> CleanupJob<F> {
-    pub fn new(f: F) -> Self {
+    pub const fn new(f: F) -> Self {
```
taurr, line 18: Nice little detail than I *const*antly forget about ;-)

src/hooks/system_mod.rs
```diff
@@ -0,0 +1,72 @@
...
+    name: &str,
+    args: rhai::Array,
+    allow_commands: bool,
+    silent: bool,
+) -> Result<Dynamic> {
+    let args: Vec<String> = args.into_iter().map(|arg| arg.to_string()).collect();
+
+    // If --allow-commands is false, we need to prompt. But we shouldn't if we're in silent mode.
+    if !allow_commands && silent {
+        return Err("Cannot prompt for system command confirmation in silent mode. Use --allow-commands if you want to allow the template to run system commands in silent mode.".into());
```
taurr, line 33: FYI: anyhow has a `bail` macro for early exit with an error

src/hooks/system_mod.rs
taurr: I don't have a hard opinion here - maybe @sassman has?

tests/integration/hooks.rs
taurr: Promts are tricky... don't think I ever solved that one either :-)

@sassman, do we have a pattern for this somewhere?

src/hooks/system_mod.rs
v-n-suadicani-issuu: This led me to realize that the Result type defined here overrides both type parameters of anyhow::Result, making it exactly the same as std::result::Result. So `bail` actually can't be used here and there is no reason to even use anyhow in this case. I unified the type definitions in a new `HookResult` type defined in the super-module that all the hook modules can use.

v-n-suadicani-issuu:
Updated to fail template generation on non-zero status from system commands.

taurr (approved):
No comments, left from me - I like it

v-n-suadicani-issuu:
@sassman Please have a look when you have time :)

tests/integration/hooks.rs
sassman: it's not really possible in an e2e test. I would address this with a very specific unit test, that makes sure that a certain function would lead to a prompt. 
For instance, with a dedicated enum variant as a return value, where the caller would then make sure to turn that value into a real prompt to the user. 
Right now we don't have such a thing yet, and prompts are quite poorly tested.

sassman (approved):
LGTM

src/hooks/system_mod.rs
sassman: not really, let's see how this flies and if users are running into issues with it.

Issue #609, opened 2022-05-29
Doc: What Regex "engine" is used


taqtiqa-mark:
Document the Regex "engine" that is used.  Otherwise it is hard to know what regex is acceptable without (painfully) iterating

sassman:
in cargo-generate we depend on the [regex crate](https://crates.io/crates/regex), there is great documentation about their support.

It might be worth to link their docs somewhere in our docs, good idea.

taqtiqa-mark:
Per this [comment](https://github.com/cargo-generate/cargo-generate/issues/618#issuecomment-1145418673) - we don't have bandwidth to help flesh out any additional functionality in this issue.

Issue #607, opened 2022-05-29
Doc: What templating language does cargo generate use?


taqtiqa-mark:
What templating language does cargo generate use?

taqtiqa-mark:
OK, this is information is buried at the end of a paragraph, half way down the templates page.

Would you accept a PR moving this information:

1. to the README.md and
2. to the [introduction page](https://cargo-generate.github.io/cargo-generate/templates/index.html)

taurr:
I for one would welcome the PR - especially if you also put a link to the scripting language `rhai`, which is also utilized as part of the templates

taqtiqa-mark:
Per this [comment](https://github.com/cargo-generate/cargo-generate/issues/618#issuecomment-1145418673) - we don't have bandwidth to help flesh out any additional functionality in this issue.

PR #610, opened 2022-05-30
Evolve[Doc]: Reference Shopify's Liquid and Rust's regex

taqtiqa-mark:
1 of 2 addressing issues #607 and #609.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,10 +14,18 @@
 
 </div>
 
-> cargo, make me a project
+```rust
+cargo generate /me/a-project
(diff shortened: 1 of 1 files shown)


sassman:
looks like a duplicate of #611 to me?

sassman:
thanks for the clarification, I've missed that.

However, I would suggest grouping all changes into 1 PR that belongs to the same semantical problem. 
So this PR and #611 are in fact addressing the same problem.

Issue #514, opened 2022-01-04
boolean value in template values file is being ignored.


morenol:
If  I try to generate using this [template](https://github.com/infinyon/fluvio-smartmodule-template/) and this as template value file:

```toml
[values]
smartmodule-type = "filter"
smartmodule-params = false
```

I get the prompt to fill the boolean parameter. If I set the boolean parameter as a string (using quotes) it works fine.

```toml
[values]
smartmodule-type = "filter"
smartmodule-params = "false"
```

PR #617, opened 2022-06-01
fix(#514): add a test to reproduce the issue

sassman:
reproduces #514


taurr:
#669 fixes this, as well as adds a test case

taurr:
Fixed by #669

PR #620, opened 2022-06-02
Generate release packages for `aarch64-unknown-linux-gnu`

SergioGasquez:
Include target `aarch64-unknown-linux-gnu` in the matrix so release packages are also generated for this target.


sassman (approved):
LGTM, thanks for your contribution

Issue #619, opened 2022-06-02
report error in windows


jm-observer:
hello, i use in windows, and it report the error:

```
PS D:\hpmq> cargo generate --git http://git.netfuse.cn/iiot-pub/hpmq-wasi-template.git
Error: 拒绝访问。 (os error 5)
```

the reason for this is https://github.com/XAMPPRocky/remove_dir_all/issues/7

i think not remove the .git folder, in turn not to copy .git to project's folder can avoid this problem?

762SPR:
I am having what I believe to be the same issue, and discussed it with some people on the Rust Embedded matrix channel who were able to replicate the issue. 

The problem seems to occur on Windows machines (I am on windows 10) when using cargo generate on a remote template. I tested with a github template [From here](https://github.com/knurling-rs/app-template)

`cargo generate 
    --git https://github.com/knurling-rs/app-template 
    --branch main 
    --name my-app`

The current workaround we found was to manually clone the template and run generate locally. Hope this helps!

sassman:
unfortunately on windows 11 for arm I was not able to reproduce this problem.

762SPR:
> I vaguely remember having that issue reproducible. The solution was very stupid, a little waiting time after the `git clone` before the recursive remove was doing the trick back then.
> 
> So [this code](https://github.com/cargo-generate/cargo-generate/blob/main/src/git/mod.rs#L148) should actually take care of the mess.
> 
> Did you actually get the warning message below?
> 
> ```rust
> warn!("Git history cleanup failed with a windows process blocking error. [Retry in {:?}]", wait_for);
> ```
> 
> Update: which cargo-generate version do you folks use?

Version was the latest, I even forced a fresh install with cargo when I was having problems.

I do recall getting some error relating to windows blocking operation where it would retry after a second or two. Sometimes it would take several retries but would always result in the permission denied/ os error 5 eventually.

This would happen maybe 20% of the time, it would usually just go straight to permission denied.

jm-observer:
i try again in other pc and it also report error. i had commited a pull(https://github.com/cargo-generate/cargo-generate/pull/621),  It should be more intuitive to look at the code

sassman:
#621 should solve this, and will be released soon

PR #621, opened 2022-06-03
Avoid panic caused by deleting git folders

jm-observer:
the reason of https://github.com/cargo-generate/cargo-generate/issues/619  is to delete _.git_ folder, so  instead of deleting _.git_, I don't copy _.git_

### DIFF
--- a/src/git/utils.rs
+++ b/src/git/utils.rs
@@ -42,9 +42,6 @@ pub fn clone_git_template_into_temp(
         .context("Please check if the Git user / repository exists.")?;
     let branch = get_branch_name_repo(&repo)?;
 
-    // change from git repository into normal folder.
-    // remove_history(git_clone_dir.path())?;
-
     Ok((git_clone_dir, branch))
 }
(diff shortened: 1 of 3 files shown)


src/git/utils.rs
```diff
@@ -43,7 +43,7 @@ pub fn clone_git_template_into_temp(
     let branch = get_branch_name_repo(&repo)?;
 
     // change from git repository into normal folder.
-    remove_history(git_clone_dir.path())?;
+    // remove_history(git_clone_dir.path())?;
```
sassman, line 46: could you please remove this code here, and also the now unused function?

sassman (changes requested):
Thanks for the neat fix, this is indeed way better, the less is copied the less needs to be deleted :)

src/git/utils.rs
jm-observer: done!

sassman (approved):
Thanks

PR #625, opened 2022-06-06
Override project_dir when using `generate()`

yozhgoor:
Hello there!

I made this PR to add the `dest` argument to the `generate` function.

This argument override the destination of the project if we give a path to the `generate` function.
If we give `None`, the behaviour stay the same as expected.

This is a breaking change btw, if you have suggestions to avoid that?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -350,37 +350,36 @@ fn locate_template_file(
     }
 }
 
+/// Resolves the project dir
+/// if `args.init == true` it either `dest` or the path of `$CWD`
 fn resolve_project_dir(name: &ProjectName, args: &Args, dest: Option<&Path>) -> Result<PathBuf> {
-    if let Some(path) = dest {
-        Ok(path.to_path_buf())
-    } else {
-        if args.init {
-            let cwd = env::current_dir()?;
-            return Ok(cwd);
-        }
+    let base_path = dest
+        .map(|p| p.to_path_buf())
+        .unwrap_or_else(|| env::current_dir().unwrap_or_else(|_| ".".into()));
 
-        let dir_name = if args.force {
-            name.raw()
-        } else {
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -350,32 +350,36 @@ fn locate_template_file(
...
-fn resolve_project_dir(name: &ProjectName, args: &Args) -> Result<PathBuf> {
-    if args.init {
-        let cwd = env::current_dir()?;
-        return Ok(cwd);
-    }
-
-    let dir_name = if args.force {
-        name.raw()
+fn resolve_project_dir(name: &ProjectName, args: &Args, dest: Option<&Path>) -> Result<PathBuf> {
+    if let Some(path) = dest {
```
sassman, line 354: I suggest avoiding the big if / else branches by leveraging default unwrapping. 
Please have a look at da42a32aead7b55828fb0bf95b154474e800be25 where I refined the function a little to be more functional style and also more DRY than it was before.

sassman (changes requested):
Nice idea, I think the breaking change is not a biggie, we can roll it with a new major version then.

sassman:
But one more note on the idea; I think this is worth to introduce as a CLI parameter, where folks can pass a different `--base-path` than `$CWD`, so if that parameter is omitted it is set to CWD. 
By doing so users can provide it via CLI params, and it would be encapsulated in `Args` so the function signature also would not have to change.

yozhgoor:
I replaced the `dest` argument from `generate()` by a `--destination` parameter. 
It conflict with `--init` so you can choose between a subfolder, the current directory or a given path.

sassman (approved):
Nice!

sassman (changes requested):
could you please also add one more test for the new argument?

PR #626, opened 2022-06-06
Update git_config to 0.4.0

taurr:
Simple update of git_config.

It was left behind at v0.2 while the newest were v0.4 - Had to change minor stuff in the code.


sassman (approved):
👍

PR #628, opened 2022-06-06
Add placeholder `{{cargo_embedded}}`.

taurr:
Enables templates to adust if they are expanded inside a
`cargo` project.

e.g. a template could leave out certain files.


sassman:
Nice idea.
So with sub project you mean something that is not a full cargo project but like a component or piece that will go into an existing project?

As for example an end to end test module? Or in general a module of a project?

taurr:
> 

In my case, in my [`bevy-template-rs` templates](http://github.com/taurr/bevy-template-rs) (it's a multi template repository), all templates can be used as stand-alone projects. Very often though, I find myself using the `crate-plugin` template for minor crates (workspace members) inside a bigger one.

In that case, the template should NOT contain all the CHANGELOG.md, LICENSE-*, cargo/rustc configuration etc. files, and I like to print a little message informing the user to remeber to add the new project to the workspace.

Also, some lines in the expanded Cargo.toml is not needed (and may even cause problems) if the current template is used in this way (for instance setting up optimization levels).

guide/src/templates/builtin_placeholders.md
```diff
@@ -0,0 +1,35 @@
...
+
+  this is supplied by either passing the `--bin` or `--lib` flag to the command line, contains either `bin` or `lib`, `--bin` is the default
+* `{{os-arch}}`
+
+  contains the current operating system and architecture ex: `linux-x86_64`
+* `{{username}}`
+
+  this will be filled in by a function borrowed from Cargo's source code, that determines your information from Cargo's configuration.
+
+* `{{cargo_embedded}}`
```
sassman, line 29: I really like the idea, but somehow the name `cargo_embedded` does not really speak to me. The terminus `embedded` is somehow very much connotated by embedded rust.

Unfortunately, I don't have a very crisp alternative to suggest.
Maybe some things that came to mind, but are probably also not good enough yet:
- `inside_a_cargo_project`
- `project_exists_already`
- `no_new_project`
- `has_existing_project`

src/lib.rs
```diff
@@ -350,10 +359,9 @@ fn locate_template_file(
     }
 }
 
-fn resolve_project_dir(name: &ProjectName, args: &Args) -> Result<PathBuf> {
+fn resolve_project_dir(base_dir: &Path, name: &ProjectName, args: &Args) -> Result<PathBuf> {
```
sassman, line 362: FYI: the parameter `base_dir` will come with #625 as a `destination` parameter as part of `Args`

guide/src/templates/builtin_placeholders.md
taurr: I agree 100% - should *not* call it anything with embedded, but that was the name that came up at the time 😄 

Of the suggestions you made, I like `inside_a_cargo_project` the most.

But there's also the plain
* `cargo_subfolder`

Another possibility is to create a `Rhai` module, maybe call it `cargo`, and create a function for the test on that. The template could still detect the situation (though it would require a scripl).

In the same direction, what it the `file` module just got a function to test if a file can be found in a parent folder? This would open up the possibility for all languages - though I admit I hate loosing the easy way that the placeholder creates for Rust. Maybe both??

taurr: Using a new `Rhai` module also opens the possibility to create functionality to test against e.g. dependencies (is Tokio available?), or to auto add dependencies etc....

But this might just be too powerfull!

sassman: I will let it sink in :)
But the tendency is right now a regular build in placeholder, not a rhai script.

taurr: Same here

barafael: Possibly, `within_cargo_project`?

src/lib.rs
taurr: I've rebased my stuff and made changes to respect the new `--destination` option, thus this will not be an issue.

guide/src/templates/builtin_placeholders.md
taurr: I liked the name @barafael suggested the most - shortest, to the point, and descriptive. That, and we talked at the Nürnberg meetup yesterday, so I'm a bit biased 😄

taurr:
@sassman any new thoughts?

guide/src/templates/builtin_placeholders.md
sassman: Sounds perfect, let's roll with this name then.

Nürnberg has a rust meetup now? How cool..

sassman:
I did oversee the comments, but I guess we have a winner name, right?

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## [Unreleased](https://github.com/cargo-generate/cargo-generate/compare/v0.14.0...HEAD)
 
+### ✨ Features
+- Add placeholder `{{within_cargo_project}}`. Enables templates to adust if they are expanded inside a `cargo` project.
```
sassman, line 6:
```suggestion
- [Add placeholder `{{within_cargo_project}}`. Enables templates to adust if they are expanded inside a `cargo` project](https://github.com/cargo-generate/cargo-generate/pull/628)
```

src/template.rs
```diff
@@ -46,9 +47,27 @@ pub fn create_liquid_object(
...
+        "within_cargo_project".into(),
+        Value::Scalar(is_within_cargo_project(project_dir).into()),
+    );
+
     Ok(liquid_object)
 }
 
+fn is_within_cargo_project(project_dir: &Path) -> bool {
+    let mut cwd = PathBuf::from(project_dir);
+    cwd.push("dummy");
```
sassman, line 60: is this needed?

sassman (changes requested):
found a minor thing

src/template.rs
taurr: yeah, `path::with_file_name()` replaces the filename in the path, so if there's no filename to begin with (as `cwd` is a path to a folder), you end up with the wrong result.

alternatively:
```rust
fn is_within_cargo_project(project_dir: &Path) -> bool {
    let mut cwd = PathBuf::from(project_dir);
    cwd.push("Cargo.toml");
    if cwd.exists() {
        return true;
    }
    while cwd.pop() && cwd.exists() {
        if cwd.with_file_name("Cargo.toml").exists() {
            return true;
        }
    }
    false
}
```

taurr: Found solution... coming up

src/template.rs
```diff
@@ -56,16 +56,8 @@ pub fn create_liquid_object(
...
-    loop {
-        if cwd.with_file_name("Cargo.toml").exists() {
-            break true;
-        }
-        if !cwd.pop() || !cwd.exists() {
-            break false;
-        }
-    }
+    let cwd = PathBuf::from(project_dir);
+    cwd.ancestors().any(|folder| folder.join("Cargo.toml").exists())
```
sassman, line 60: `cwd.ancestors()` would go up, up until the root folder. 

We need a way to stop at the "current project" boundaries, otherwise this would yield unexpected results.

sassman (changes requested):

src/template.rs
sassman: I see your point.

But I can also imagine cases where that would lead to surprises. Anyways let's ship it and wait for complaints :)

taurr: I even suspect this is exactly what `cargo` itself does in order to find the project root if you do a `cargo build` some random place

taurr: lol, I actually expected this behaviour of cargo - how else should it resolve for workspaces?

sassman (approved):

PR #651, opened 2022-06-15
`cargo generate` without options panics

### LINKED ISSUE #650: `cargo generate` without options panics
After the change to `Clap`, `cargo-generate` fails in panic instead of printing the help if invoked without parameters.

**To Reproduce**
1. invoke `cargo generate`.
2. Output:
```shell
~ ❱ cargo generate
💡   Using application config: /home/johnny/.cargo/cargo-generate.toml
thread 'main' panicked at 'assertion failed: args.favorite.is_some()', /home/johnny/.cargo/registry/src/github.com-1ecc6299db9ec823/cargo-generate-0.15.0/src/user_parsed_input.rs:79:9
```

**Expected behavior**
I expected `cargo-generate` to output its help message.

 - OS: Linux && Windows
 - `cargo generate --version`: 0.15.0

taurr:
Fixes #650

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -4,19 +4,25 @@ use std::{
 };
 
 use anyhow::{anyhow, Result};
-use clap::Parser;
+use clap::{ArgGroup, Args, Parser};
 
 use crate::git;
 
 #[derive(Parser)]
+#[clap(name = "cargo")]
 #[clap(bin_name = "cargo")]
 pub enum Cli {
     #[clap(name = "generate", visible_alias = "gen")]
-    Generate(Args),
+    Generate(GenerateArgs),
 }
 
-#[derive(Clone, Debug, Parser)]
-pub struct Args {
+#[derive(Clone, Debug, Args)]
+#[clap(group(
+    ArgGroup::new("template")
+        .required(true)
+        .args(&["favorite", "git", "path", "list-favorites"]),
+))]
+pub struct GenerateArgs {
     /// List defined favorite templates from the config
     #[clap(
         long,
--- a/src/favorites.rs
+++ b/src/favorites.rs
@@ -2,12 +2,12 @@
 
 use crate::{
     app_config::{AppConfig, FavoriteConfig},
-    emoji, Args,
+    emoji, GenerateArgs,
 };
 use anyhow::Result;
 use console::style;
 
-pub fn list_favorites(app_config: &AppConfig, args: &Args) -> Result<()> {
+pub fn list_favorites(app_config: &AppConfig, args: &GenerateArgs) -> Result<()> {
     let data = {
         let mut d = app_config
             .favorites
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -74,7 +74,7 @@ use crate::{
 };
 
 /// # Panics
-pub fn generate(mut args: Args) -> Result<()> {
+pub fn generate(mut args: GenerateArgs) -> Result<()> {
     let app_config: AppConfig = app_config_path(&args.config)?.as_path().try_into()?;
 
     if args.list_favorites {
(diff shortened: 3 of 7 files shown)


src/args.rs
```diff
@@ -4,19 +4,25 @@ use std::{
 };
 
 use anyhow::{anyhow, Result};
-use clap::Parser;
+use clap::{ArgGroup, Args, Parser};
 
 use crate::git;
 
 #[derive(Parser)]
+#[clap(name = "cargo")]
```
sassman, line 12:
```suggestion
#[clap(name = "cargo generate", arg_required_else_help(true))]
```

src/args.rs
```diff
@@ -4,19 +4,25 @@ use std::{
...
 #[clap(bin_name = "cargo")]
```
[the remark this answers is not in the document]
sassman: this will re-add the `-V | --version` argument, which is gone right now and it broke the last release asset generation job.

src/args.rs
```diff
@@ -4,19 +4,26 @@ use std::{
...
 
 use anyhow::{anyhow, Result};
-use clap::Parser;
+use clap::{ArgGroup, Args, Parser};
 
 use crate::git;
 
 #[derive(Parser)]
+#[clap(name = "cargo generate", arg_required_else_help(true))]
+#[clap(version)]
```
sassman, line 13:
```suggestion
#[clap(version)]
#[clap(about)]
```

sassman (approved):
thanks for fixing that. I added few more commits for the issues I have observed during the last release.

src/args.rs
```diff
@@ -147,6 +111,83 @@ pub struct GenerateArgs {
...
+        self.git
+            .as_ref()
+            .or(self.path.as_ref())
+            .or(self.favorite.as_ref())
+            .or(self.auto_path.as_ref())
+            .cloned()
+            .unwrap()
+    }
+
+    pub fn git(&self) -> Option<String> {
```
sassman, line 162: nit: It would be enough, and more resource-friendly to expose here a `&str`. Let the caller decide if they need to take ownership of a clone.

sassman: The same goes for the other accessor methods

sassman:
> Also, if we want to introduce subcommands (and we do if we are to fix introduction of a test command as specified), it'll be a breaking change... IMHO we should cleanup this mess and introduce subcmds asap. How big of a breaking change? I don't know atm. but I can try and come up with a suggestion, then work from there.

Let's work on this on a follow-up PR and Issue. 
I plan to release soon a patch version.

src/args.rs
```diff
@@ -149,41 +149,40 @@ impl TemplatePath {
...
             .or(self.path.as_ref())
             .or(self.favorite.as_ref())
             .or(self.auto_path.as_ref())
-            .cloned()
             .unwrap()
     }
 
-    pub fn git(&self) -> Option<String> {
-        self.git.clone()
+    pub const fn git(&self) -> Option<&String> {
```
sassman, line 161: nit: I would prefer to go with `&str` like `self.git.as_str()`

taurr: nopez.... to do that requires use of `.map(|s| s.as_ref())`, and then there's issues on the calling side because it actually call other fn's that expect `Option<&String>`

Could then make it into `Cow<>` but is it worth it?

sassman: it's possible to annotate the lifetime, you can find an [example here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e1de171af6a240a85f5a26c88ff7d414)

sassman: let me put a shameless plug here, my latest blog post on `AsRef<>`:
https://d34dl0ck.me/rust-bites-as-ref/index.html

taurr: we're actually in [this situation](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b599397c6bf258f3c0a50a44ddc6bb43).

As I said, I don't know whether it's worth it - its way harder to read, and all the callsites actually expect `Option<String>`, so to really gain anything, we should also change that code... now it starts to grow into its own PR.

sassman: It's ok, we don't need to go down this rabbit hole

taurr: @sassman, as stated, don't know whether it was worth it...

sassman (approved):

PR #661, opened 2022-06-19
feat: new `Rhai` variable: `is_init` ?

### LINKED ISSUE #649: feat: new `Rhai` variable: `is_init` ?
In line with the new `within_cargo_project` variable, a template may need to adjust to the `--init` option.

Specifically, if a template is expanded with `--init`, then it may follow that certain files is not needed. This could be files like `LICENSE-MIT` or `CHANGELOG.md`.

taurr:
Fixes #649

### DIFF
--- a/src/template.rs
+++ b/src/template.rs
@@ -8,7 +8,6 @@ use std::path::Path;
 use walkdir::{DirEntry, WalkDir};
 
 use crate::config::TemplateConfig;
-use crate::emoji;
 use crate::filenames::substitute_filename;
 use crate::include_exclude::*;
 use crate::progressbar::spinner;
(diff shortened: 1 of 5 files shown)


src/template.rs
```diff
@@ -27,14 +27,18 @@ fn engine() -> liquid::Parser {
...
+    args: &GenerateArgs,
     project_dir: &Path,
     name: &ProjectName,
     crate_type: &CrateType,
-    force: bool,
 ) -> Result<Object> {
     let authors: Authors = get_authors()?;
     let os_arch = get_os_arch();
-    let project_name = if force { name.raw() } else { name.kebab_case() };
+    let project_name = if args.force {
```
sassman, line 37: Nit: do you know the `.then()` method on bool? It's a beautiful way to handle this thing..

sassman (approved):

src/template.rs
taurr: fixed.

I knew it, but didn't think to change the line because:
* it was already there
* I don't think `then()` on bool is especially readable

anyhow - its done!

PR #666, opened 2022-06-25
Return the path of the generated project from the `generate` function

yozhgoor:
Hello there!

This PR aims to return the path of the generated project from the `generate` function.
Allowing the user of the public API to re-use the path when they can't guess the folder name of the generated project.

The changes there are quite big for this not so big goal, I you have any suggestions to make it clearer.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,10 +1,16 @@
 use anyhow::Result;
-use cargo_generate::{generate, Cli};
+use cargo_generate::{app_config_path, generate, list_favorites, AppConfig, Cli};
 use clap::Parser;
 
 fn main() -> Result<()> {
     let Cli::Generate(args) = Cli::parse();
-    generate(args)?;
+    let app_config: AppConfig = app_config_path(&args.config)?.as_path().try_into()?;
+
+    if args.list_favorites {
+        list_favorites(&app_config, &args)?;
+    } else {
+        generate(app_config, args)?;
+    }
 
     Ok(())
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,15 +1,14 @@
 use anyhow::Result;
-use cargo_generate::{app_config_path, generate, list_favorites, AppConfig, Cli};
+use cargo_generate::{generate, list_favorites, Cli};
 use clap::Parser;
 
 fn main() -> Result<()> {
     let Cli::Generate(args) = Cli::parse();
-    let app_config: AppConfig = app_config_path(&args.config)?.as_path().try_into()?;
 
     if args.list_favorites {
-        list_favorites(&app_config, &args)?;
+        list_favorites(&args)?;
     } else {
-        generate(app_config, args)?;
+        generate(args)?;
     }
 
     Ok(())
--- a/tests/integration/helpers/project.rs
+++ b/tests/integration/helpers/project.rs
@@ -41,13 +41,3 @@ impl Project {
         self.path().join(path).exists()
     }
 }
-
-pub fn read(path: &Path) -> String {
-    let mut ret = String::new();
-    File::open(&path)
-        .unwrap_or_else(|_| panic!("couldn't open file {:?}", path))
-        .read_to_string(&mut ret)
-        .unwrap_or_else(|_| panic!("couldn't read file {:?}", path));
-
-    ret
-}
--- a/tests/integration/helpers/project.rs
+++ b/tests/integration/helpers/project.rs
@@ -41,3 +41,13 @@ impl Project {
         self.path().join(path).exists()
     }
 }
+
+pub fn read(path: &Path) -> String {
+    let mut ret = String::new();
+    File::open(&path)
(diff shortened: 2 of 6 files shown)


src/main.rs
```diff
@@ -1,10 +1,16 @@
 use anyhow::Result;
-use cargo_generate::{generate, Cli};
+use cargo_generate::{app_config_path, generate, list_favorites, AppConfig, Cli};
 use clap::Parser;
 
 fn main() -> Result<()> {
     let Cli::Generate(args) = Cli::parse();
-    generate(args)?;
+    let app_config: AppConfig = app_config_path(&args.config)?.as_path().try_into()?;
```
taurr, line 7: This is a nice little refactoring if you ask me...
Makes it more clearcut between the `list-favorites` and normal `generate` functionality..

tests/integration/library.rs
```diff
@@ -43,9 +43,16 @@ version = "0.1.0"
...
+    let app_config: AppConfig = app_config_path(&args_exposed.config)
+        .expect("cannot get app's config path")
+        .as_path()
+        .try_into()
+        .expect("cannot parse app's config path");
+
     // need to cd to the dir as we aren't running in the cargo shell.
     assert!(std::env::set_current_dir(&dir.root).is_ok());
-    assert!(generate(args_exposed).is_ok());
+    assert!(generate(app_config, args_exposed).is_ok());
```
taurr, line 55: I would also look at the return value here - there's no test on the `Result<PathBuf>`, which was the point of this refactoring...

taurr (changes requested):
In general; I like it - but **why** do you need the path to be returned?
It's not used nor explained anywhere - not even in the tests.

yozhgoor:
> In general; I like it - but **why** do you need the path to be returned? It's not used nor explained anywhere - not even in the tests.

My use case is that I want to create the template in a temporary directory. But I can't guess the generated directory and using `GenerateArgs.destination` directly means that I'm a directory level to far (`base/tmp_dir/template` instead of `base/template`). The idea is to return the path directly from the `generate` so I can re-use it directly so I can add files after creating the template.

yozhgoor:
I reduced to a single test, but I can split them into two different test if you prefer that.

sassman (approved):
LGTM, this separation is a very good idea.

sassman:
only a fix for the test is required:

```plain
test library::it_allows_generate_call_with_public_args_and_returns_generated_path ... FAILED
failures:
---- library::it_allows_generate_call_with_public_args_and_returns_generated_path stdout ----
🔧   Basedir: /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/.tmpf8jHPV ...
🔧   Generating template ...
🔧   Moving generated files into: `/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/.tmpf8jHPV/foobar_project`...
💡   Initializing a fresh Git repository
✨   Done! New project created /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/.tmpf8jHPV/foobar_project
thread 'library::it_allows_generate_call_with_public_args_and_returns_generated_path' panicked at 'assertion failed: `(left == right)`
  left: `"/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/.tmpf8jHPV/foobar_project"`,
 right: `"/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/.tmpf8jHPV/foobar_project"`', tests/integration/library.rs:49:5
```

tests/integration/library.rs
sassman: I think the recently added test is fine, it verifies the returned path

yozhgoor:
> only a fix for the test is required:

I'm on it, not really sure how I can fix this. This is really weird to me since I use the same variable for the two paths.

yozhgoor:
I'm not proud of the tests, I need to investigate a bit more.
I'm gonna draft for the moment (need to go to the kid's fancy fair :dancers:).

yozhgoor:
> macOS creates several folders of caches and temporary items to speed up the system. Those under your control are located in /Library/Caches and /private/tmp. But system-related cache files are hidden in the /private/var folder on Mac.

Seems linked to the caching system from macOS, even my dirty try who added the `/private` part of the path by myself doesn't work.

taurr (approved):

tests/integration/helpers/project.rs
```diff
@@ -41,3 +41,13 @@ impl Project {
         self.path().join(path).exists()
     }
 }
+
+pub fn read(path: &Path) -> String {
```
sassman, line 45: do you know that there is [`fs::read_to_string()`](https://doc.rust-lang.org/std/fs/fn.read_to_string.html) doing exactly this?

yozhgoor: Yeah, the `read` function is completely useless here

Issue #667, opened 2022-06-25
Add `.deb` package release


John15321:
I can see that you already have an arch package version of this project added which is mentioned in your guide. A `.deb` package can easily be maid using this crate:
https://crates.io/crates/cargo-deb

Im working on proving a similar release artifact for another crate, so I thought that I may offer you the same. What do you think?

sassman:
Sure, any contribution is welcome.

John15321:
@sassman The PR is up 😄 please check it out

PR #668, opened 2022-06-26
#667 add creating and release a deb package

John15321:
Add a CI job for creating a `.deb` package using the `cargo-deb` crate and adds it to the release.

Please check that the workflow does not destroy other releasing artifacts

Resolves: https://github.com/cargo-generate/cargo-generate/issues/667


.github/workflows/deb-release.yml
```diff
@@ -0,0 +1,37 @@
+name: Release deb package
```
sassman, line 1: could you please either refactor that flow into this file https://github.com/cargo-generate/cargo-generate/blob/main/.github/workflows/release-binary-assets.yml
as a dedicated job, 

or change the trigger that it's triggered from that named release-binary-assets workflow.

John15321: sure i will try to add it to the `release-binary-assets.yml` workflow

John15321:
@sassman Hey im trying to work on that but for some reason, the pipeline started failing due to some linting issue which is weird because I haven't touched anything in the source code. Could you please take a look at that?
https://github.com/John15321/cargo-generate/actions/runs/2600909326

.github/workflows/release-binary-assets.yml
```diff
@@ -55,6 +55,18 @@ jobs:
           command: run
           use-cross: ${{ matrix.cross }}
           args: --features vendored-openssl --release --target=${{ matrix.target }} -- --version
+      - name: Install cargo-deb and create a deb package
+        if: matrix.os == 'ubuntu-latest'
+        run: |
+          cargo install cargo-deb
+          cargo deb
```
sassman, line 62: could you please figure out if there is a good way how to pass over the args 

```
  --features vendored-openssl --release --target=${{ matrix.target }} -- --version
```

so that we make sure the generated deb file does use the `vendored-openssl` feautre and has a matching target architecture, otherwise we would generate here multiple deb packages for linux. see the matrix configuration above.

sassman (changes requested):

.github/workflows/release-binary-assets.yml
John15321: @sassman Sorry I wasnt available is this still an issue? I can see that the PR is merged

sassman: @John15321 I believe it is an issue and the PR is still open :)

John15321: > @John15321 I believe it is an issue and the PR is still open :)



> @John15321 I believe it is an issue and the PR is still open :)

Oh sorry, maybe its better to not reply just after waking up lol

PR #669, opened 2022-06-26
fix(#514): boolean value being ignored

### LINKED ISSUE #514: boolean value in template values file is being ignored.

taurr:
Fixes #514

Also fixes/enables use of placeholders in multiple levels


src/lib.rs
```diff
@@ -467,14 +465,21 @@ fn expand_template(
     Ok(())
 }
 
+/// Try to add all provided template_values to the liquid_object.
+///
+/// ## Note:
+/// Values for which a placeholder exists, should already be filled by `fill_project_variables`
 pub(crate) fn add_missing_provided_values(
-    mut liquid_object: liquid::Object,
+    liquid_object: Rc<RefCell<liquid::Object>>,
```
sassman, line 473: why would we need and want to introduce a `Rc<RefCell>` here?
I think so far we made a good job avoiding any smart pointers, let's keep that spirit as hard as we can.

taurr: That's simple my friend; I was lazy, and didn't want to clone due to the `'static` requirements from the `rhai` api.

fix comming up!

src/hooks/mod.rs
```diff
@@ -76,6 +66,31 @@ fn evaluate_scripts(dir: &Path, scripts: &[String], engine: rhai::Engine) -> Res
     Ok(())
 }
 
+pub fn evaluate_script<T: Clone + 'static>(
+    liquid_object: liquid::Object,
+    script: &str,
+) -> Result<T, Box<rhai::EvalAltResult>> {
+    let mut conditional_evaluation_engine = rhai::Engine::new();
+
+    #[allow(deprecated)]
```
sassman, line 75: Is there not yet way out of this deprecation? 
I'm a bit hesitant to let new functionality use deprecated features.

Can you provide a little context on why it make sense to use this very deprecated feature?

src/lib.rs
```diff
@@ -486,86 +481,93 @@ pub(crate) fn add_missing_provided_values(
...
-    Ok(liquid_object)
+    Ok(())
 }
 
-fn merge_conditionals(
-    template_config: &Config,
-    liquid_object: liquid::Object,
+fn fill_placeholders_and_merge_conditionals(
+    config: &mut Config,
+    mut liquid_object: liquid::Object,
```
sassman, line 489: `liquid_object` is now mutable, so why are we still need to return a `liquid::Object` then?

With a mutable input we could set and fill all the things we need, without returning a new `Object`

src/project_variables.rs
```diff
@@ -92,41 +95,38 @@ const RESERVED_NAMES: [&str; 7] = [
 ];
 
 pub fn fill_project_variables<F>(
-    mut template_object: Object,
-    template_config: &Config,
+    template_object: &mut Object,
```
sassman, line 98: reference to the previous comment.

Here it's done exactly this way, mutable input, no return `Object` anylonger.

src/hooks/mod.rs
taurr: not that I know of.

It is not a new one either - it's been there since we introduced `rhai`. They deprecatec the api because it *might* change according to their documentation.

src/lib.rs
taurr: because we need to take ownership!

The `Rhai` API needs a `Rc<Cell<Object>> + 'static` - thus we need to pass ownership whenever we need the engine to resove stuff. It's either that (like we did before my changes here), convert the var into Rc<T> like I did originally in this PR, or resorting to `unsafe` which I wont do.

Or you can try your hand at introducing the proper lifetime specs in `Rhai` - I gave up the other day - tried but gave up as it would introduce major API breakning which doubt they want.

src/project_variables.rs
taurr: but here, the API does *not* need the ownership

src/lib.rs
taurr: @sassman ?

sassman: I See, yeah lifetime hell is also unpleasant.

sassman (approved):
🚀

PR #677, opened 2022-07-03
Feature request: Set default VCS from within templates

### LINKED ISSUE #635: Feature request: Set default VCS from within templates
It would be nice if it was possible to make templates default to a different or no VCS initialisation. In my case this would be useful for a "minimal" template meant for messing around inside. Could also be useful for templates which are meant to be a smaller part of a larger vcs repository.

As a workaround I tried to delete the initialised vcs repository from rhai, but that doesn't work as rhai scripts run before vcs initalisation.

taurr:
Fixes #635


sassman (approved):
LGTM

PR #686, opened 2022-07-05
Add the `current_year` placeholder

yozhgoor:
Hello there, this PR aims to add the `current_year` placeholder.

I seen that the `it_subsitute_date()` test exists but I didn't seen much documentation about that. This placeholder will replace `{{current_year}}` with the current year using [chrono::Local](https://docs.rs/chrono/latest/chrono/offset/struct.Local.html) instead of a formatted given date.


yozhgoor:
My bad, TIL about Shopify's [liquid language](https://shopify.github.io/liquid/filters/date/).

PR #691, opened 2022-07-07
Add option to overwrite files in existing project

### LINKED ISSUE #569: Add option to overwrite files in existing project
[The book has the following to say](https://cargo-generate.github.io/cargo-generate/usage.html#generating-into-current-dir) about this:

> `cargo-generate` will not allow any existing files to be overwritten and will fail to generate any files should there be any conflicts.

I actually think this would be a good feature to have, although the default choice of refusing to overwrite is good. Perhaps an `--overwrite` flag could be added to generate over the existing project anyway?

For an example use case, I have [this template](https://or.computer.surgery/charles/nix-rust-quickstart) that receives relatively frequent updates since Nix Flakes are still evolving. I feel like an `--overwrite` option would make it easier to propagate these changes and new best practices to existing projects generated from this template. Then one could do `cargo gen ... --overwrite && git restore -p .` instead of having to figure out how to apply all of the changes manually.

taurr:
Fixes #569


sassman (approved):
LGTM 🚀

PR #693, opened 2022-07-09
Allow the template author to manually specify sub-temlates

### LINKED ISSUE #636: Feat: when pointing to a multi-template repo, it should be possible to explicitly specify the default template
When using `cargo-generate` with a multi template repository, `cargo-generate` creates a `choice` prompt for the user to select a 
template to expand.

It would help on usability if it was possible to influence this prompt.

1) It should be possible to specify the default selected template.
2) It would be nice if the template author could specify the sorting order, or specify the list of templates manually.

taurr:
Closes #636

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -210,29 +210,24 @@ fn get_source_template_into_temp(
     Ok((temp_dir, branch))
 }
 
-fn resolve_project_name(args: &GenerateArgs) -> Result<ProjectName> {
-    match args.name {
-        Some(ref n) => Ok(ProjectName::new(n)),
-        None if !args.silent => Ok(ProjectName::new(interactive::name()?)),
-        None => Err(anyhow!(
-            "{} {} {}",
-            emoji::ERROR,
-            style("Project Name Error:").bold().red(),
-            style("Option `--silent` provided, but project name was not set. Please use `--name`.")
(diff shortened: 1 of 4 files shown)


taurr:
With this change, sub-templates may also be "nested" within each other

src/lib.rs
```diff
@@ -259,41 +254,103 @@ fn resolve_template_dir(template_base_dir: &TempDir, subfolder: Option<&str>) ->
...
                 var_name: "Template".into(),
                 var_info: VarInfo::String {
                     entry: Box::new(StringEntry {
-                        default: Some(config_paths[0].clone()),
-                        choices: Some(config_paths),
+                        default: Some(config_paths[0].display().to_string()),
+                        choices: Some(
+                            config_paths
+                                .into_iter()
+                                .map(|p| p.display().to_string())
```
sassman, line 290: Maybe it's a good idea to verify here if the file or directory does exist, before asking the user?

sassman (approved):

src/lib.rs
taurr: @sassman, not really needed, as the `locate_template_configs` function only returns paths with a `cargo-generate.toml` file in them - thus ensuring "valid" template paths.

PR #699, opened 2022-07-14
Support `--test` for running tests on the expanded template

### LINKED ISSUE #455: `cargo test` shortcut for template authors

taurr:
This is an attempt at supporting `--test` for running `cargo --test` inside the temporary folder where the template is expanded.

The implementation is now 100% in line with the original feature request.

* Anything after the `--test` on the command line is forwarded to the `cargo test` command.
* If needed, the user can specify a `CARGO_GENERATE_TEST_CMD` environment variable, whose contents will be used instead of `cargo test`, thereby making the command work for non Rust templates.
* User does not have to specify a project-name, as one will be generated.
* `--test` can not be used with `--git`, `--path` or `--favorite`. If will *always* use `--path .` internally.
* If a favorite/subfolder is specified, it will be interpreted as a subfolder of $CWD.

t-rec_2

Fixes #455

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -147,60 +147,85 @@ fn internal_generate(mut args: GenerateArgs) -> Result<PathBuf> {
     )?;
 
     if args.template_path.test {
-        println!(
-            "{} {}{}{}",
-            emoji::WRENCH,
-            style("Running \"").bold(),
-            style("cargo test"),
-            style("\" ...").bold(),
-        );
-        std::env::set_current_dir(&template_dir)?;
-        let (cmd, cmd_args) = std::env::var("CARGO_GENERATE_TEST_CMD")
-            .map(|env_test_cmd| {
-                let mut split_cmd_args = env_test_cmd.split_whitespace().map(str::to_string);
-                (
-                    split_cmd_args.next().unwrap(),
-                    split_cmd_args.collect::<Vec<String>>(),
-                )
-            })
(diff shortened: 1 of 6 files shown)


taurr:
Still missing automatic tests...

src/lib.rs
```diff
@@ -164,18 +184,50 @@ fn internal_generate(mut args: GenerateArgs) -> Result<PathBuf> {
...
-
     Ok(project_dir)
 }
 
+fn test_expanded_template(template_dir: &PathBuf, args: &GenerateArgs) -> Result<PathBuf> {
+    println!(
+        "{} {}{}{}",
+        emoji::WRENCH,
+        style("Running \"").bold(),
+        style("cargo test"),
```
sassman, line 202: here `cargo test` is hard coded, but below the actual command is taken from `std::env::var("CARGO_GENERATE_TEST_CMD")` with the default of `cargo test`.

IMO it would be better to first extract the command that will be used finally, and reuse this command here in the `println!`.

sassman: we can tackle this as a follow up..

PR #707, opened 2022-07-26
Removed Serenity Template from Template List

chilipepperhott:
Hey I haven't maintained that template in months... possibly years?

So I removed it from the list. I hope to avoid wasting anyone else's time with it.


sassman (approved):
Thanks 🙏

Issue #708, opened 2022-07-26
Release Binary Assets failed for release `0.16.0`


SergioGasquez:
[CI for releasing assets for `0.16.0` failed](https://github.com/cargo-generate/cargo-generate/runs/7499419924?check_suite_focus=true) with:
```bash
Error: The process '/home/runner/.cargo/bin/cross' failed with exit code 2
```

SergioGasquez:
I am not sure if this is something you can fix on your end, or if it's related to `cross` but in any case, we could comment the smoke test to rerun the actions and have the binaries available for the failing architectures.

sassman:
Looks a bit like `cross` accept arguments now somehow different than before. Not sure but a manual trigger would be a good idea as well.

SergioGasquez:
> Looks a bit like `cross` accept arguments now somehow different than before. Not sure but a manual trigger would be a good idea as well.

Running the same commands on my local machine does not throw the error that we are seeing in the CI:
```
❯ cross build --features vendored-openssl --release --target=x86_64-unknown-linux-musl
    Finished release [optimized] target(s) in 0.14s

❯ cross run --features vendored-openssl --release --target=x86_64-unknown-linux-musl -- --version
    Finished release [optimized] target(s) in 0.11s
     Running `/qemu-runner x86_64 /target/x86_64-unknown-linux-musl/release/cargo-generate`
cargo generate 0.16.0
cargo, make me a project

USAGE:
    cargo <SUBCOMMAND>

OPTIONS:
    -h, --help       Print help information
    -V, --version    Print version information

SUBCOMMANDS:
    generate    cargo, make me a project [aliases: gen]
    help        Print this message or the help of the given subcommand(s)
```

jessebraham:
Why is `cross` being used at all? AFAIK all of these failing targets can be built on `ubuntu-latest` without it.

SergioGasquez:
Hello @sassman, do you have any updates regarding this issue? Would you accept a PR commenting the `smoke test` step just to get the release binaries generated? Our CI uses the binaries and it's currently failing because of this issue.

SergioGasquez:
Hello @sassman! Since #721 got merged, could we rerun the `Release Binary Assets` action and close this issue?

SergioGasquez:
Hello @sassman, sorry for insisting but, do you have any updates regarding this issue?

PR #721, opened 2022-08-08
Remove `Smoke Test` step

SergioGasquez:
As discussed in #708, this step was failing and causing the `Release Binary Assets` action to fail. I've commented it out in the meantime to relaunch the action and generate binaries.


sassman (approved):

SergioGasquez:
Thanks for merging it so fast! Can you rerun https://github.com/cargo-generate/cargo-generate/actions/runs/2732136316 to generate the artefacts?

PR #778, opened 2022-11-05
Fix submodule test for git 2.38.1 and similar

### LINKED ISSUE #777: git 2.38.1 breaks basics::it_loads_a_submodule test
Most of the template doesn't apply since this bug was hit while packing cargo-generate not while using it.

**Describe the bug**
In environment with git 2.38.1 test ` basics::it_loads_a_submodule` seems to fail. See [info about git release](https://www.phoronix.com/news/Git-2.38.1-Released). I think patches to the mentioned CVEs are likely the culprit.

```
failures:

---- basics::it_loads_a_submodule stdout ----
thread 'basics::it_loads_a_submodule' panicked at 'Unexpected failure.
code-128
stderr=```"Cloning into \'/build/.tmpWg7xJ7/submodule\'...\nfatal: transport \'file\' not allowed\nfatal: clone of \'file:///build/.tmptH4CUn\' into submodule path \'/build/.tmpWg7xJ7/submodule\' failed\n"```
command=`"git" "submodule" "add" "file:///build/.tmptH4CUn" "./submodule/"`
code=128
stdout=""
stderr="Cloning into \'/build/.tmpWg7xJ7/submodule\'...\nfatal: transport \'file\' not allowed\nfatal: clone of \'file:///build/.tmptH4CUn\' into submodule path \'/build/.tmpWg7xJ7/submodule\' failed\n"
', /build/rustc-1.64.0-src/library/core/src/ops/function.rs:248:5
```

**make sure you**
- [x] ~I'm on the latest rust version `rustup update`~ yes but using nix instead of rustup
- [x] ~I'm on the latest cargo-generate version `cargo install cargo-generate`~ installed via nix

**To Reproduce**
1. `cargo test` on system with git 2.38.1
2. above

**Expected behavior**
Tests should pass

**Screenshots**

 - OS: Linux
 - Version: `Linux nixos-mainframe 5.15.75 #1-NixOS SMP Wed Oct 26 10:35:57 UTC 2022 x86_64 GNU/Linux`
 - Build tools: gnu
 - Shell: zsh & bash
 - `rustc --version`: 1.64.0
 - `cargo generate --version`: 0.16.0

**Additional context**
downstream fix https://github.com/NixOS/nixpkgs/pull/199668

turboMaCk:
resolves #777

Recent fixes of CVE-2022-39253 and CVE-2022-39253 in git broke submodule integration test.

This fix adds configuration option to enable necessary operations.

__I've tested this patch on Linux x86_64 with both git 2.38.1 and 2.38.0.__

### DIFF
--- a/tests/integration/helpers/project_builder.rs
+++ b/tests/integration/helpers/project_builder.rs
@@ -157,6 +157,8 @@ version = "0.1.0"
 
             self.submodules.iter().for_each(|(d, m)| {
                 Command::new("git")
+                    .arg("-c")
+                    .arg("protocol.file.allow=always")
                     .arg("submodule")
                     .arg("add")
                     .arg(&m)


sassman:
Thanks for that fix.
Would you mind also checking the clippy lints? That would be awesome 🤩

turboMaCk:
@sassman hi! I actually already tried when I saw the steps failing. Anyway I was not able to reproduce them locally.

Since they look unrelated to the change made [see output](https://github.com/cargo-generate/cargo-generate/actions/runs/3399787301/jobs/5653792886) I simply gave up on looking deeper.

My theory is that this is upstream change. Likely CI started installing different version of clippy which made these new issues surfaced. They seems to be mostly mechanical things - removing pattern matching and replacing it with functions for Option and similar. I would be even willing to contribute the fix for these but first I need to find a way to reproduce it locally since I would probably go mad if I had to do it based on feedback from CI.

I'll a bit more time looking into how I can install the version clippy is using but can't promise anything. There are quite a few more things on my plate for today.

turboMaCk:
The clippy related errors seems to be caused by update from `1.64.0` to `1.65.0` stable toolchain. It seems someone patched rustup to work on NixOS so I'm able to use to to install same versions CI does and reproduce error.

I'll be opening separate PR for this to keep history after merges clear.

PR #779, opened 2022-11-05
Fix clippy after 1.65.0 toolchain update

turboMaCk:
Addressing https://github.com/cargo-generate/cargo-generate/pull/778#issuecomment-1304503386

As a side note - I found it quite hard to contribute to the project because of amount of assumptions (which are not true for my setup) it makes.

- I'm using `commit.gpgsign true` in my git config and tests are picking up the global config prompting me for passphrase and making tests failed as they don't expect signing of git commits.
- I find git-hooks particularly annoying especially in combination with issues like one above. I pretty much just disabled them with `--no-verify` to be able to make any progress.

I believe that decisions like this are up to the projects maintainers to make so don't take this as criticism just a data about workflow of other people which might not have been considered. Personally I found it quite annoying since the workarounds feel like downgrade from the workflow I normally use (signing all commits or being able to run tests in specific environment I'm testing). I think the assumption that environment in which commits and pushes are made is the same environment for which code is written (and thus tests are written) for is problematic. Especially since there would be value for project like this to be able to test against specific versions of git and with various combinations of git configurations. Current philosophy though is exact opposite of that. Not only that tests depend on global git configuration but also git commit hooks enforce that even the environment from which `git push` is being made is the one in which tests pass. While in my case I'm forced to run tests as a different user to the user I'm committing as.

### DIFF
--- a/src/git/mod.rs
+++ b/src/git/mod.rs
@@ -148,16 +148,16 @@ pub fn init(project_dir: &Path, branch: Option<&str>, force: bool) -> Git2Result
         Repository::init_opts(project_dir, &opts)
     }
 
-    match Repository::discover(project_dir) {
-        Ok(repo) => {
+    Repository::discover(project_dir).map_or_else(
+        |_| just_init(project_dir, branch),
+        |repo| {
             if force {
                 Repository::open(project_dir).or_else(|_| just_init(project_dir, branch))
             } else {
                 Ok(repo)
             }
-        }
-        Err(_) => just_init(project_dir, branch),
-    }
+        },
+    )
 }
 
 /// remove context of repository by removing `.git` from filesystem
--- a/src/git/utils.rs
+++ b/src/git/utils.rs
@@ -46,10 +46,10 @@ pub fn clone_git_template_into_temp(
     if let Some(tag) = tag {
         let (object, reference) = repo.revparse_ext(tag)?;
         repo.checkout_tree(&object, None)?;
-        match reference {
-            Some(gref) => repo.set_head(gref.name().unwrap()),
-            None => repo.set_head_detached(object.id()),
-        }?;
+        reference.map_or_else(
+            || repo.set_head_detached(object.id()),
+            |gref| repo.set_head(gref.name().unwrap()),
+        )?
     }
 
     Ok((git_clone_dir, branch))
--- a/src/hooks/mod.rs
+++ b/src/hooks/mod.rs
@@ -75,17 +75,18 @@ pub fn evaluate_script<T: Clone + 'static>(
 
     #[allow(deprecated)]
     conditional_evaluation_engine.on_var({
-        move |name, _, _| match liquid_object.get(name) {
-            Some(value) => Ok(value.as_view().as_scalar().map(|scalar| {
-                scalar.to_bool().map_or_else(
-                    || {
-                        let v = scalar.to_kstr();
-                        v.as_str().into()
-                    },
-                    |v| v.into(),
-                )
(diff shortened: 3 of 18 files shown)


turboMaCk:
Failure in this branch is what https://github.com/cargo-generate/cargo-generate/pull/778#issuecomment-1304503386 fixes.

So full test fix is combination of these 2 PRs. However each one addresses different set of issues so I'm keeping them separate. Feel free to merge them to one before merge to main if you prefer.

sassman:
> Addressing https://github.com/cargo-generate/cargo-generate/pull/778#issuecomment-1304503386
> 
> 
> 
> As a side note - I found it quite hard to contribute to the project because of amount of assumptions (which are not true for my setup) it makes.
> 
> 
> 
> - I'm using `commit.gpgsign true` in my git config and tests are picking up the global config prompting me for passphrase and making tests failed as they don't expect signing of git commits.
> 
> - I find git-hooks particularly annoying especially in combination with issues like one above. I pretty much just disabled them with `--no-verify` to be able to make any progress.
> 
> 
> 
> I believe that decisions like this are up to the projects maintainers to make so don't take this as criticism just a data about workflow of other people which might not have been considered. Personally I found it quite annoying since the workarounds feel like downgrade from the workflow I normally use (signing all commits or being able to run tests in specific environment I'm testing). I think the assumption that environment in which commits and pushes are made is the same environment for which code is written (and thus tests are written) for is problematic. Especially since there would be value for project like this to be able to test against specific versions of git and with various combinations of git configurations. Current philosophy though is exact opposite of that. Not only that tests depend on global git configuration but also git commit hooks enforce that even the environment from which `git push` is being made is the one in which tests pass. While in my case I'm forced to run tests as a different user to the user I'm committing as.
> 
> 

Those are very valuable points. I'll have a look if we can change the tests in a way that the made assumptions are set explicitly for the test runs. 
So far I've not encountered the problems myself, even with commit signing that I use by myself as well.

Anyhow thanks for the feedback those things will be addressed.

turboMaCk:
It might be more related to my setup of GPG. I'm also using it as ssh agent. In both cases I have time based expiration after which I'm prompted for passphrase again.

The root of the problem though is that git simply looks into all the configuration levels up to the $HOME. Maybe the possible strategy would be to run tests in environment where `HOME` is set to some directory like `tests/HOME` which would avoid `~/.gitconfig` from changing the behavior of tests? I'm not really sure but it might be simple to test. For now I just used workarounds to temporarily with combination of temporarily disabling signing in my global gitconfig and `--no-verify` flags. If I have a need to contribute further patches to the project I will consider exploring the options to make this work more smoothly for myself without breaking existing workflows.

sassman:
> It might be more related to my setup of GPG. I'm also using it as an ssh agent. In both cases I have time based expiration after which I'm prompted for passphrase again.
> 
> 
> 
> The root of the problem though is that git simply looks into all the configuration levels up to the $HOME. Maybe the possible strategy would be to run tests in environment where `HOME` is set to some directory like `tests/HOME` which would avoid `~/.gitconfig` from changing the behavior of tests? I'm not really sure but it might be simple to test. For now I just used workarounds like combination of temporarily disabling signing in my global gitconfig and `--no-verify` flags. If I have a need to contribute further patches to the project I will consider exploring the options to make this work more smoothly for myself without breaking existing workflows.

Yeah, I had something something similar in mind. I thought we create a .gitignore in the test folder so that it is used by the integration tests. By doing so it would be independent of the user's configuration, so side effects shouldn't happen.

PR #781, opened 2022-11-06
When using --init, do not copy .git directories

MalloryA:
Fixes: https://github.com/cargo-generate/cargo-generate/issues/758


sassman (approved):
LGTM 🚀

Issue #834, opened 2022-12-27
Building with Nix


nothingnesses:
I'd like to be able to build cargo-generate with the Nix package manager so I can integrate it with other projects built with the latter. Please could some support for this use case be added?

PR #835, opened 2022-12-27
Add Nix flake

nothingnesses:
Should address #834.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,6 @@
+(import (let lock = builtins.fromJSON (builtins.readFile ./flake.lock);
+in fetchTarball {
+  url =
+    "https://github.com/edolstra/flake-compat/archive/${lock.nodes.flake-compat.locked.rev}.tar.gz";
+  sha256 = lock.nodes.flake-compat.locked.narHash;
+}) { src = ./.; }).defaultNix
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,222 @@
+{
+  "nodes": {
+    "crane": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1670900067,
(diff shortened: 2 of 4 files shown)


default.nix
```diff
@@ -0,0 +1,6 @@
+(import (let lock = builtins.fromJSON (builtins.readFile ./flake.lock);
+in fetchTarball {
+  url =
+    "https://github.com/edolstra/flake-compat/archive/${lock.nodes.flake-compat.locked.rev}.tar.gz";
```
sassman, line 4: can you please elaborate on what URL that is, who maintains that, and how can we be sure this is safe to use?

nothingnesses: Sure thing. That's the URL for the flake-compat library which allows projects using the newer Flakes mechanism (i.e., what I'm adding with this PR) [to be used with derivations that don't use Flakes](https://nixos.wiki/wiki/Flakes#Using_flakes_project_from_a_legacy_Nix). It's maintained by Eelco Dolstra who wrote [the thesis that gave birth to the Nix package manager](https://edolstra.github.io/pubs/phd-thesis.pdf). The [source code for the library](https://github.com/edolstra/flake-compat/blob/master/default.nix) is somewhat small (200 lines), so it's feasible to review it quite quickly for anything malicious.

neosam:
I personally think it would make more sense to put those files in a separate repository.  Instead of using `src=./.`, you could use `src=fetchFromGitHub` to check out the cargo-generate's source code.

nothingnesses:
@neosam Ah, I didn't know that was possible. I'll have to try that out, thanks.

neosam:
> @neosam Ah, I didn't know that was possible. I'll have to try that out, thanks.

You're very welcome.  Not sure if you need the latest version of cargo-generate but there is already cargo-generate in Nix. https://search.nixos.org/packages?channel=22.11&show=cargo-generate&from=0&size=50&sort=relevance&type=packages&query=cargo-generate

PR #844, opened 2023-01-07
show a warning to the user if the template doesn't agree with the cli parameters

### LINKED ISSUE #763: show a warning to the user if the template doesn't agree with the cli parameters
in lib.rs we have a TODO in this - it quite relevant, so we should probably se if we can do something about it.

taurr:
Fixes #763

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -98,12 +98,21 @@ fn internal_generate(args: GenerateArgs) -> Result<PathBuf> {
     )?;
 
     // the `--init` parameter may also be set by the template itself
-    // TODO: should we show a warning to the user if the template doesn't agree with the parameter? The user might not expect the template files to end up in CWD!
-    user_parsed_input.init |= config
+    if config
         .template
         .as_ref()
         .and_then(|c| c.init)
-        .unwrap_or(false);
+        .unwrap_or(false)
+        && !user_parsed_input.init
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -98,12 +98,21 @@ fn internal_generate(args: GenerateArgs) -> Result<PathBuf> {
...
-    user_parsed_input.init |= config
+    if config
         .template
         .as_ref()
         .and_then(|c| c.init)
-        .unwrap_or(false);
+        .unwrap_or(false)
+        && !user_parsed_input.init
+    {
+        eprintln!(
```
sassman, line 108: would it not make sense to prompt the user with a question here, so that he can actively consent to the `--init` behavior or reject it and stop hereby the template processing?

taurr: I for one would be kinda upset if `cargo-generate` did that to me - the reason for specifying the init behavior in the template is because the template is only designed to use with that behavior, and the user is too lazy to specify it on the command line every time (or too forgetful in my case!).

We've had the TODO in the code for quite a while, to just warn the user that this is happening - so the only behavior changed is the actual warning.

sassman: also very reasonable, fine with me.

sassman (approved):

PR #856, opened 2023-01-19
build: put vendored libraries behind features

figsoda:
this makes it easier for downstream repos to use the system libraries without having to patch cargo-generate

`vendored-openssl` was previously not a default feature, but since we were using `git/vendored-openssl` unconditionally, it was still enabled by the default


sassman:
@figsoda Thanks for your contribution, could you please rebase your PR and also document the feature flags in the `README.md` and in the guide (`guide/src/installation.md`)

sassman (approved):
LGTM

figsoda:
thanks!

Issue #855, opened 2023-01-19
Feat: New `Liquid` filter for running a `Rhai` script


taurr:
Based on a discussion in #847, the idea came up to have a `Liquid` filter to allow executing `Rhai` scripts and include the output in the template.

I'm currently thinking on a syntax like 

```liquid
Example execution of a Rhai script: {{script_file|rhai}}
```

Caveat: at `Liquid` filters only takes 1 parameter, it's rather hard to pass anything to the `Rhai` scripts without misusing the syntax in some way - which is why I am not proposing parameters for the `Rhai` script. Ideas or alternatives more than welcome.

dmitris:
nit on Feat: `s/Fea:/Feat: ` in the Subject 😄

Issue #847, opened 2023-01-10
is it possible to set a variable shared between different files?


dmitris:
Is it possible to set a variable that is not a placeholder? I'd like to set a variable to an array of strings and then use it in conditionals.  Something like:
```
{% assign program_type_with_opts = "xdp, classifier, sock_ops, cgroup_skb, cgroup_sysctl, cgroup_sockopt, uprobe, uretprobe" | split: ", " %}
{% if program_type in program_type_with_opts %}
use clap::Parser;
{% endif %}
```
I saw the `assign` in https://shopify.github.io/liquid/tags/variable/ but was not able to use it in a cargo-generate template.

If this is not possible currently, would you consider an extension allowing this?  It would help to avoid "ugly" workarounds like in https://github.com/aya-rs/aya-template/blob/2ba73f6f2f8b24f15d3d197878191258e7d7d397/%7B%7Bproject-name%7D%7D/src/main.rs#L47-L48 and https://github.com/aya-rs/aya-template/blob/2ba73f6f2f8b24f15d3d197878191258e7d7d397/%7B%7Bproject-name%7D%7D/src/main.rs#L67-L68 😄  Thanks!

Related to https://github.com/aya-rs/aya-template/pull/67 and https://github.com/aya-rs/aya-template/pull/68.

dmitris:
I now figured out how to use the `assign` - therefore modified the subject - and was able to do the following:
```
{% assign program_types_with_opts = "xdp, classifier, sock_ops, cgroup_skb, cgroup_sysctl, cgroup_sockopt, uprobe, uretprobe" | split: ", " %}

{% if program_types_with_opts contains program_type %}
use clap::Parser;
{% endif %}
```

Now I need to figure out how to assign a variable in one file and use it in another in the DRY spirit, to avoid doing it in several places (and risking the later edits getting out of syncs, or having to update the same assignment in multiple places).  Do you know how / where to do it?  `cargo-generate.toml` would be a good place for that - but at least for now, I wasn't able to put the assignment there, would appreciate any hint 😄

taurr:
It is very much possible - just use the variable extension as described in [the book](https://cargo-generate.github.io/cargo-generate/templates/scripting.rhai-extensions.html).

You cannot use the Rhai assign keyword for this, as Rhai sees each file as its own, and not as a project - thus **there's NO carryover between files. Nor is there a specific order!!!**

[...]

PR #866, opened 2023-02-01
Taurr/issue_#855_Feat-New-`Liquid`-filter-for-running-a-`Rhai`-script

taurr:
Make it possible to use a rhai script to supply the value for a liquid substitution.

Works as a simple filter using the syntax:
```liquid
Execute a rhai script as a filter {{"script_file"|rhai}}
```

One can even use liquid variables to supply the script filename.

@sassman: I'm know tests & documentation are missing (working on it), but could you provide some feedback pls.?

Running on the new example, the output in the CLI is:
```shell
💡   Using application config: /home/johnny/.cargo/cargo-generate.toml
🔧   Destination: /home/johnny/Projects/github/cargo-generate/test-template-expanded/t1 ...
🔧   project-name: t1 ...
🔧   Generating template ...
[1/2] ⠁ Processing: hello.txt                                                                                                  Inside the rhai filter function
Inside the rhai filter function
[1/2]   Done: hello.txt                                                                                                        
[2/2]   Skipped: rhai-example.rhai - used as Rhai filter!                                                                      🔧   Moving generated files into: `/home/johnny/Projects/github/cargo-generate/test-template-expanded/t1`...
💡   Initializing a fresh Git repository
✨   Done! New project created /home/johnny/Projects/github/cargo-generate/test-template-expanded/t1
```

and the template output (a single file) is:
```txt
This is an example of how to use the VERY cool rhai filter feature.
We use a Liquid filter to run a Rhai script:

This is the result of running the script as a filter.

You can even set values from within the rhai filter function

Though after liquid itself has used the value, or processing the next file has begun, any changes
to the variable by a rhai filter will be ignored (by liquid) until the next liquid file!

Any liquid variable created/assigned in a file is not carried over to other files, but the ones from Rhai are!
```


example-templates/liquid-rhai-bridge/rhai-example.rhai
```diff
@@ -0,0 +1,5 @@
+print("Inside the rhai filter function");
+
+variable::set("some_liquid_variable", "You can even set values from within the rhai filter function");
```
sassman, line 3: is the expectation that those variables set in rhai exists over the whole lifetime of cargo generate or only for this very rhai script?

taurr: Definitely the whole - variables not designated as placeholders but assigned from `liquid` only last for the file in which they are defined, while these last from definition and onwards. This is consistent with the behavior if doing it in a hook function.

Come to think about it, it would be nice if we could somehow enforce the order of processing files. That would simplify some of the scripts a user might write (thinking of #847 here)

taurr:
@dmitris: could you have a look at this? I think you'll be interested!

taurr:
@sassman Only a squash and a merge left - if it can get approved!

dmitris (commented):
👍 LGTM, thanks!

taurr:
closes #855

Issue #899, opened 2023-03-17
Replace println! with logging for API users of cargo-generate


mothran:
**Describe the bug**
Currently cargo-generate prints message to the user via println / eprintln. This works well when using it as a CLI tool but if a user wants to extend the tool via the API from: https://docs.rs/cargo-generate/0.18.1/cargo_generate/ or wrap it. It would be nice to be able to redirect or silence the logging.

**Expected behavior**
Via the cargo-generate cli entrypoint I would expect registering a env_logger or tracing_subscriber that then wrote to stdout, but when calling https://docs.rs/cargo-generate/0.18.1/cargo_generate/fn.generate.html or other methods exposed by the API, allow the caller to define the logging behavior.

I would be happy to provide a PR if the team has a preferred logging engine (ex: `log` vs `tracing`)

Thanks!

sassman:
Sounds reasonable, `log` and `env_logger` should be sufficient for this use case.

PR #907, opened 2023-03-24
Migrate from println! to log+env_logger

mothran:
Fixes: https://github.com/cargo-generate/cargo-generate/issues/899

* Removed log.rs
* replaced all eprintln/println calls to log::* calls
* connected the verbose flag to toggle ProgressBar draw calls

Currently the env_logger defaults to replicate the println! behavior. I did my best to translate the println -> info and eprintln -> warn. But I am sure the levels of different messages might be wrong in a few places.


mothran:
Looks like both of the CI failures are related to TLS cert validation. It might need to be manually re-run by a contributor or org owner.

src/git/mod.rs
```diff
@@ -175,7 +175,7 @@ pub fn remove_history(project_dir: &Path) -> io::Result<()> {
 
                 if e.to_string().contains("The process cannot access the file because it is being used by another process.") {
                     let wait_for = Duration::from_secs(2_u64.pow(attempt.sub(1).into()));
-                    warn!("Git history cleanup failed with a windows process blocking error. [Retry in {:?}]", wait_for);
+                    warn!("{} Git history cleanup failed with a windows process blocking error. [Retry in {:?}]", emoji::WARN, wait_for);
```
sassman, line 178: Instead of prefixing the emoji like this, I recommend to configure the env_logger in a way that each log level would include those prefixes.

You can see an [example in the rust cookbook](https://rust-lang-nursery.github.io/rust-cookbook/development_tools/debugging/config_log.html#include-timestamp-in-log-messages)

mothran: Good point! I know there is a WARN emoji but how would I map other log levels to the defined emoji's here?:
https://github.com/cargo-generate/cargo-generate/blob/main/src/emoji.rs#L5
because there are times when info is used with non-info like sparkle:
https://github.com/cargo-generate/cargo-generate/blob/main/src/favorites.rs#L55

sassman: Maybe it's a good opportunity to align them. Or if that's too noisy maybe only warning and error are good candidates for now.

mothran: Great, I will take that just integrates warn|error emoji levels at the formatter.

mothran: First shot at addressing it here: cf8f6e6e70b596876a67e4bc58817b16781f7cfc

mothran: @sassman is there anything else I should address?

sassman: thanks for addressing this! Looks very good so far.

src/template.rs
```diff
@@ -316,8 +319,7 @@ pub fn render_string_gracefully(
 
 fn print_files_with_errors_warning(files_with_errors: Vec<(String, liquid_core::Error)>) {
     let mut msg = format!(
-        "\n{} {}",
-        emoji::WARN,
+        "\n{}",
```
sassman, line 322: I believe this does now behave unintended.

Before the change it was printed a `\n<warn-emoji> <message>`
Now it's going to be `<warn-emoji>\n<message>`

suggestion: removing the `\n` completely here and therefore logging an `info!("\n")` at just before line 336. That should lead to the same behavior as before.

sassman (approved):
LGTM 🚀 
Thanks a lot for your contribution and going the last mile.

mothran:
@sassman thank you! Appreciate the feedback and thanks for the great project.

PR #916, opened 2023-04-11
fix: `.liquid` suffixes not stripped when cloning git templates

KaitlynEthylia:
When creating the temporary directory from a path template, the .git folder was removed and .liquid suffixes were stripped from files. This did not happen with git templates, which caused a disparity in how the `ignore` field was handled. This implements this when cloning the git repo, thus fixing the disparity


sassman (approved):
LGTM 🚀 
Thanks for your contribution. Very much appreciated.

PR #917, opened 2023-04-11
Update `clap` and `tempfile`

yozhgoor:
Update to match dependencies of `cargo-temp`


sassman:
I bumped the version with the commit 1c6496a9f63e3db8a8b91b13ff83c336438b54f0 already. Thanks for providing this fix.

PR #930, opened 2023-04-25
fix: fixes prompt order (#885)

### LINKED ISSUE #885: Define a prompt order
Hi! Would it be possible to make the order in which the prompts appear deterministic? At the moment the order is "random" every time.  I think it's possible to move them into a RHAI script, but then our several CIs that generate the template would stop working as we use the `-d` option to choose from the prompts.

sassman:
This PR fixes #885 

- the prompt order is now as defined in the config file
- re-arranges also the tests a bit


sassman:
The final implementation now echos a value of a template variable when provided via the CLI, that was the only way to have a valid test covering the new feature.

image

PR #936, opened 2023-05-04
Add ".toml" to default config file reference

sshine:


guide/src/templates/custom_placeholders.md
```diff
@@ -143,7 +143,7 @@ For the file format, see above.
 
 #### Default values
 
-Default values may be specified in the config file (specified with the `--config` flag, or in the default config file `$CARGO_HOME/cargo-generate`)
+Default values may be specified in the config file (specified with the `--config` flag, or in the default config file `$CARGO_HOME/cargo-generate.toml`)
```
sassman, line 146: actually, both files are possible. Or in other words the extension `.toml` is possible but optional.

sshine: I wouldn’t have guessed. :)

Do you think either one is preferable?

stappersg:
Those we can close this stalled merge request, please do.  It prevents that this dangling MR will drain further human energy.

PR #955, opened 2023-06-01
Fix link to 'ignoring files'

sshine:
Presumably this used to be a section in include_exclude.md, but is now its own page.


sassman (approved):
good catch, thanks a lot.

PR #978, opened 2023-07-05
fix: use tempdir prefix

jonaro00:
In rare cases, if a generate operation is forcefully cancelled/panics, the tempdir used for git cloning might not be cleared up.
The default `.tmp` prefix leaves the tempdir hidden and hard to know which program created it.
Having `cargo-generate` prefix adds clarity and visibility.

The same is true when running tests, for instance if you cancel the pre-push hook.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -133,6 +133,7 @@ pub fn locate_template_configs(base_dir: &Path) -> Result<Vec<PathBuf>> {
 #[cfg(test)]
 mod tests {
     use crate::tests::create_file;
+    use crate::tmp_dir;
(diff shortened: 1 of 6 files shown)


sassman (commented):
Thanks for your PR.

But I see now a lot of code duplications. 

I would appreciate if you could refactor a bit and DRY the code duplications and move the `tmpdir` function into the utility module `utils.rs` and start using it from all over the place.

jonaro00:
Done

sassman (approved):
LGTM 🚀

PR #1014, opened 2023-08-31
feat: support abbreviation for sourcehut

sgeisenh:


README.md
```diff
@@ -51,6 +51,7 @@ cargo generate username-on-github/mytemplate
 # templates on other git platforms
 cargo generate gl:username-on-gitlab/mytemplate
 cargo generate bb:username-on-bitbucket/mytemplate
+cargo generate sr:~username-on-sourcehut/mytemplate
```
sassman, line 54: I do understand that sourcehut uses `~` as a username prefix. 
However, running this on the shell might cause problems, because the tilde (`~`) represents the home directory of the currently logged in user.

So I would suggest (event this might be counter intuitive) to implement it in the way that a user would only name the username without the tilde and cargo-generate does add the tilde behind the scenes.

e.g. this:

```
cargo generate sr:username-without-tilde-on-sourcehut/mytemplate`
```

translates into this:

```
cargo generate --git 'https://git.sr.ht/~username-without-tilde-on-sourcehut/mytemplate'
```

I believe it's important that the whole url is in single quotes otherwhise the shell might interpret the tilde as a home dir path. (not verified though)

sgeisenh: I adjusted the behavior so that the tilde is added by cargo-generate. I didn't wrap the generated URL in single quotes because as far as I can tell, everything internally is handled by calls to git2. Does that seem reasonable?

sassman: Absolutely! Thanks a lot.

I would recommend adding a note to the docs to elaborate on this behavior, that users might not expect.

sgeisenh: A added some details to `README.md` and `guide/src/usage.md` describing the expansion behavior (and drawing special attention to the additional tilde for sourcehut).

sassman (approved):
Thanks a lot for going the extra mile

Issue #1020, opened 2023-09-01
Tracking Issue for SSH related problems


sassman:
## List of related issues

- http(s) to ssh rewrite: 
  - #874 
- protected ssh keys: 
  - #817
- default ssh key assumption: 
  - #384 
- tests for ssh repo
  - #383

sassman:
All fixed issues will be released with `v0.22.0`

Issue #817, opened 2022-12-08
`cargo generate` command does nothing


rakuten-JO:
**Describe the bug**
I want to start a project:
```shell
cargo generate --git https://github.com/rustwasm/wasm-pack-template.git
```
And I get:
```shell
Error: Please check if the Git user / repository exists.

Caused by:
    Failed to retrieve list of SSH authentication methods: Failed getting response; class=Ssh (23); code=Auth (-16)
```

**Desktop (please complete the following information):**
 - OS: [MacOS]
 - Version: [Mac Os Big Sur version 11.4]
 - `rustc --version`: rustc 1.65.0 (897e37553 2022-11-02)
 - `cargo generate --version`: cargo generate-generate 0.17.3

**Additional context**
it is fine in my Windows PC, but i guess it may be the problem of my .git config.
```shell
# Personal account, - the default config
Host github.com-personal
   HostName github.com
   User git
   IdentityFile ~/.ssh/id_rsa
   
# Work account
Host github.com-organization
   HostName github.com
   User git
   IdentityFile ~/.ssh/id_ed25519
```
And I want to know how to fix it?

sassman:
Interesting, are your keys protected by a passphrase?

As it seems it is using git over ssh even if you've given it a http url.

When you comment out the .gitconfig lines you've mentioned then I guess it does work, right?

sassman:
@rakuten-JO can you please try the approach from the comment below and see if that would solve your issue?

https://github.com/rust-lang/cargo/issues/3381#issuecomment-725980194

moridinga:
> @rakuten-JO can you please try the approach from the comment below and see if that would solve your issue?
> 
> [rust-lang/cargo#3381 (comment)](https://github.com/rust-lang/cargo/issues/3381#issuecomment-725980194)

I've tried that approach (`git-fetch-with-cli = true` in the config.toml for cargo), and it doesn't work for cargo generate. I would have expected it to, but it does not.

stappersg:
Augustus 2023, it works for me ...
```text
stappers@paddy:/tmp
$ cargo generate --git https://github.com/rustwasm/wasm-pack-template.git
🤷   Project Name: foo
🔧   Destination: /tmp/foo ...
🔧   project-name: foo ...
🔧   Generating template ...
🔧   Moving generated files into: `/tmp/foo`...
Initializing a fresh Git repository
✨   Done! New project created /tmp/foo
stappers@paddy:/tmp
[...]

PR #1025, opened 2023-09-06
Use auth-git2 for git authentication

de-vri-es:
This PR uses `auth-git2` for git authentication. It supports much more authentication methods out of the box.

It will try to use the SSH agent, default key files for multiple key types in `$HOME/.ssh`, the git credentials helper and falls back to asking the user for authentication using the correct askpess helper if configured, or just prompt on the terminal if available.

For private key files, it can also automatically ask the user for the password if it is an encrypted OpenSSH key (also using askpass and falling back to prompting on the terminal).

It can be customized to disable any of the options. For example, the prompting on the terminal could be disabled if not desired.

Additionally, the second commit sets the fetch options when updating submodules too, so that authentication is also used for those.


src/git/mod.rs
```diff
@@ -71,64 +64,75 @@ impl<'cb> RepoCloneBuilder<'cb> {
...
+        log::info!(
+            "{} `{}` {}",
+            style("Using private key:").bold(),
+            style(format_args!("{}", identity_path.display()))
+                .bold()
+                .yellow(),
+            style("for git-ssh checkout").bold()
+        );
+        self.authenticator =
+            GitAuthenticator::new_empty().add_ssh_key_from_file(identity_path, None);
```
sassman, line 84:
```suggestion
            GitAuthenticator::new_empty().add_ssh_key_from_file(identity_path, None)            .prompt_ssh_key_password(true);
```

This change would support passphrase protected ssh key files, because it would still ask for the passphrase if needed.

de-vri-es: Ah, good point. I forgot about enabling that :)

Cargo.toml
```diff
@@ -11,6 +11,7 @@ include = ["src/**/*", "LICENSE-*", "*.md"]
 [dependencies]
 clap = { version = "~4.4", features = ["derive", "std", "help"], default-features = false }
 git2 = { version = "~0.18", features = ["ssh", "https"], default-features = false }
+auth-git2 = "0.5.0"
```
sassman, line 14:
```suggestion
auth-git2 = "~0.5"
```

We're used to pin the dependencies that way within the project.

sassman:
This PR is related to #1020 and will solve 
- #817 
- #384

sassman (approved):
Thanks a lot for your contribution, that issue was a long standing and quite painful thing. Glad to have it sorted out now.

de-vri-es:
Happy to contribute! Thanks for the review :D

PR #1055, opened 2023-11-10
feat(error): use `fs_err` instead of `std::fs` to provide actionable error messages

LukeMathWalker:
I.e. messages that include the file or directory that caused the error.

I was building a new template using `cargo-generate` today and I got stuck on a "File name is too long" error. No way to find out what file was causing the issue, even by passing `--verbose` to the CLI. Hence this PR!

After the change, I was able to immediately pinpoint the file and get unblocked.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -146,6 +146,7 @@ dependencies = [
  "console",
  "dialoguer",
  "env_logger",
+ "fs-err",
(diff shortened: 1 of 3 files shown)


sassman (approved):
Amazing, had no idea `fs_err` exists. Loving it.

Issue #1111, opened 2024-01-24
feature: make submodule downloads optional


SchmErik:
Hello! We've been using this crate in a command line utility to generate templates from our project and we've run in to issues with slow template generation. From what I understand `cargo-generate` is typically used for generating templates from repos. However, we use it on a template that is within a [mono repo](https://github.com/risc0/risc0/tree/main/templates/rust-starter) with submodules that are not necessary for the template. The template is kept in this repo to facilitate testing various components and we find that this model helps all of the parts stay in sync until we solidify our APIs.

We've created a command line utility that uses the `cargo-generate` crate functions to generate a template. In our scenario, the submodule download can add significant overhead during template generation due to downloading unnecessary submodules. Some of our users have reported running our template generation command for tens of minutes. We would like cargo generate crate functions allow ignoring submodules.

I have a draft implementation of this locally (currently trying to diagnose a failing test case) so if there aren't objections, I'd like to open a PR!

PR #1112, opened 2024-01-24
add --skip-submodules flag to optionalize cloning git submodules

SchmErik:
`cargo-generate` is typically used for generating templates from small repos. However, we would like to use it on [our repo] where the template lies in a larger mono-repo with git submodules that are not necessary to the template. The template is kept in this repo to failitate testing the various components within this repo. In this situation, the submodule download can add significat overhead to the initialization of the repo. Some have reported running our template generation command for tens of minutes and impacts the usability of our tool.

This change makes submodule cloning optional by adding a `--skip-submodules` flag that callers can set when using cargo-generate crate functions. This change intends the default behavior to clone submodules if they exist.

[our repo]: https://github.com/risc0/risc0

Fixes: #1111

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -121,7 +121,7 @@ pub struct GenerateArgs {
     #[arg(short, long, action)]
     pub overwrite: bool,
 
-    /// Download git submodules (if there are any)
+    /// Skip downloading git submodules (if there are any)
     #[arg(long, action)]
     pub skip_submodules: bool,
 
--- a/src/args.rs
+++ b/src/args.rs
@@ -121,6 +121,10 @@ pub struct GenerateArgs {
     #[arg(short, long, action)]
     pub overwrite: bool,
 
+    /// Download git submodules (if there are any)
+    #[arg(long, action)]
+    pub skip_submodules: bool,
+
     /// All args after "--" on the command line.
     #[arg(skip)]
     pub other_args: Option<Vec<String>>,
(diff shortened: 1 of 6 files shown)


src/args.rs
```diff
@@ -121,6 +121,10 @@ pub struct GenerateArgs {
     #[arg(short, long, action)]
     pub overwrite: bool,
 
+    /// Download git submodules (if there are any)
```
sassman, line 124:
```suggestion
    /// Skip download git submodules (if there are any)
```

this was presumably meant, right?

src/git/mod.rs
```diff
@@ -23,24 +23,26 @@ pub use utils::{tmp_dir, try_get_branch_from_path};
 
 // Assumptions:
 // * `--git <url>` should only be parse in the same way as `git clone <url>` would
-// * submodules should be clone by default
+// * submodules can be cloned by setting the submodules field to true.
```
sassman, line 26: Are you certain the text is actually right?

I have understood that `--skip-submodules=false` will lead to checked out submodules (the default case) 
and `--skip-submodules=true` will not check them out.

Or do I miss anything?

sassman:
Thanks for your contribution. 
It would be nice to have the submodule related tests extended to cover the new skip flag.

src/git/mod.rs
sassman, line 26:
```suggestion
// * submodules are cloned by default, but can be skipped by `--skip-submodules`.
```

sassman (approved):

PR #1211, opened 2024-06-24
fix: actually set `--depth 1`

MingweiSamuel:
On large repos cargo-generate takes several minutes to run. This lowers that to... a dozen or so seconds, which is still pretty bad


sassman (approved):

sassman:
Thanks for your PR! Great idea.

MingweiSamuel:
Thanks for the merge!

Issue #894, opened 2023-03-07
Simplify / shorten / reorganize `cargo generate`'s help text.


alilleybrinker:
Currently, the help text for `cargo generate` is quite large, both vertically (there are many flags) and horizontally (the descriptions for the flags are quite long).

On my own system I have:

<details>
    <summary>Click to view the help text.</summary>

```
cargo, make me a project

Usage: 

Arguments:
  [AUTO_PATH]
          Auto attempt to use as either `--git` or `--favorite`. If either is specified explicitly, use as subfolder

  [SUBFOLDER]
          Specifies a subfolder within the template repository to be used as the actual template

Options:
      --test
          Expand $CWD as a template, then run `cargo test` on the expansion (set $CARGO_GENERATE_TEST_CMD to override test command).
          
          Any arguments given after the `--test` argument, will be used as arguments for the test command.

  -g, --git <GIT>
          Git repository to clone template from. Can be a URL (like `https://github.com/rust-cli/cli-template`), a path (relative or absolute), or an `owner/repo` abbreviated GitHub URL (like `rust-cli/cli-template`).
          
          Note that cargo generate will first attempt to interpret the `owner/repo` form as a relative path and only try a GitHub URL if the local path doesn't exist.

  -b, --branch <BRANCH>
          Branch to use when installing from git

  -t, --tag <TAG>
          Tag to use when installing from git

  -p, --path <PATH>
          Local path to copy the template from. Can not be specified together with --git

      --favorite <FAVORITE>
          Generate a favorite template as defined in the config. In case the favorite is undefined, use in place of the `--git` option, otherwise specifies the subfolder

      --list-favorites
          List defined favorite templates from the config

  -n, --name <NAME>
          Directory to create / project name; if the name isn't in kebab-case, it will be converted to kebab-case unless `--force` is given

  -f, --force
          Don't convert the project name to kebab-case before creating the directory. Note that cargo generate won't overwrite an existing directory, even if `--force` is given

  -v, --verbose
          Enables more verbose output

      --template-values-file <TEMPLATE_VALUES_FILE>
          Pass template values through a file Values should be in the format `key=value`, one per line

  -s, --silent
          If silent mode is set all variables will be extracted from the template_values_file. If a value is missing the project generation will fail

  -c, --config <CONFIG>
          Use specific configuration file. Defaults to $CARGO_HOME/cargo-generate or $HOME/.cargo/cargo-generate

      --vcs <VCS>
[...]

PR #1223, opened 2024-07-15
feat: reformat help text in the style of cargo

joshka:
My motivation here was that when I run cargo generate, I find it difficult to find the options that are useful as there's no real order to the help output. This helps by injecting color, sorting the options into some reasonable sections, and shortening the longest option name.

# Before:

image
image

# After:

image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,12 +12,7 @@ include = ["src/**/*", "LICENSE-*", "*.md"]
 anstyle = "1.0.7"
 anyhow = "~1.0"
 auth-git2 = "~0.5"
-clap = { version = "~4.5", default-features = true, features = [
-    "derive",
-    "std",
-    "help",
-    "wrap_help",
-] }
+clap = { version = "~4.5", features = ["derive", "wrap_help"] }
 console = "~0.15"
 dialoguer = "~0.11"
 env_logger = "~0.11"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,58 +9,54 @@ edition = "2021"
 include = ["src/**/*", "LICENSE-*", "*.md"]
 
 [dependencies]
-clap = { version = "~4.5", features = ["derive", "std", "help"], default-features = false }
-git2 = { version = "~0.19", features = ["ssh", "https"], default-features = false }
+anyhow = "~1.0"
 auth-git2 = "~0.5"
+clap = { version = "~4.5", default-features = false, features = [
+    "derive",
+    "std",
+    "help",
+] }
 console = "~0.15"
 dialoguer = "~0.11"
-indicatif = "~0.17"
-tempfile = "~3.10"
-regex = "~1.10"
-heck = "~0.5"
-walkdir = "~2.5"
-remove_dir_all = "~0.8"
-ignore = "~0.4"
-anyhow = "~1.0"
-time = "~0.3"
-toml = { version = "~0.8", features = ["preserve_order"] }
-thiserror = "~1.0"
-home = "~0.5"
-sanitize-filename = "~0.5"
-rhai = "~1.19"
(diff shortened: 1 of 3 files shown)


Cargo.toml
joshka, line 1: The formatting / sorting changes are in a separate commit

Cargo.toml
```diff
@@ -9,58 +9,56 @@ edition = "2021"
...
-git2 = { version = "~0.19", features = ["ssh", "https"], default-features = false }
+anstyle = "1.0.7"
+anyhow = "~1.0"
 auth-git2 = "~0.5"
+clap = { version = "~4.5", default-features = true, features = [
+    "derive",
+    "std",
+    "help",
+    "wrap_help",
+] }
```
joshka, line 20: This is the only real change in this file (adding wrap help and turning on default features, which activates the color and the usage features)

src/args.rs
```diff
@@ -9,13 +9,47 @@ use serde::Deserialize;
...
+            .valid(VALID)
+            .invalid(INVALID)
+            .error(ERROR)
+    };
+}
+
+mod heading {
+    pub const GIT_PARAMETERS: &str = "Git Parameters";
+    pub const TEMPLATE_SELECTION: &str = "Template Selection";
+    pub const OUTPUT_PARAMETERS: &str = "Output Parameters";
```
joshka, line 41: I wasn't 100% sure of good names for these headings, and there's a few options that I wasn't entirely sure whether to leave them in the main options section or in the output params (e.g. overwrite, allow-commands)

src/args.rs
```diff
@@ -52,77 +86,78 @@ pub struct GenerateArgs {
...
     /// Enables more verbose output.
     #[arg(long, short, action)]
     pub verbose: bool,
 
-    /// Pass template values through a file
-    /// Values should be in the format `key=value`, one per line
-    #[arg(long, value_parser)]
+    /// Pass template values through a file. Values should be in the format `key=value`, one per
+    /// line
+    #[arg(long="values-file", value_parser, alias="template-values-file", value_name="FILE", help_heading = heading::OUTPUT_PARAMETERS)]
```
joshka, line 103: Added a shorter name here to make the formatting nicer (and also to make this easier for a user to type). I left the original alias however so any existing scripts etc. can still use that name.

Cargo.toml
```diff
@@ -9,58 +9,51 @@ edition = "2021"
 include = ["src/**/*", "LICENSE-*", "*.md"]
 
 [dependencies]
-clap = { version = "~4.5", features = ["derive", "std", "help"], default-features = false }
-git2 = { version = "~0.19", features = ["ssh", "https"], default-features = false }
+anstyle = "1.0.7"
+anyhow = "~1.0"
 auth-git2 = "~0.5"
+clap = { version = "~4.5", features = ["derive", "wrap_help"] }
```
joshka, line 15: This is the only real change in this file (adding wrap help and turning on default features, which activates the color and the usage features).

There's no specific reason in the git / PR history why default features was turned off here for clap. See https://github.com/cargo-generate/cargo-generate/pull/655

src/args.rs
sassman: for me it sounds reasonable.

src/args.rs
sassman: thanks for keeping it as a non-breaking change.

sassman (approved):
Thanks a lot for this huge improvement! I know that several folks already asked for this.

sassman:
this PR addresses parts of #894

Cargo.toml
sassman: fair point.

joshka:
Thanks for the quick merge :)

PR #1284, opened 2024-10-02
chore(deps): don't request ineffective ssh_key_from_memory feature from git2-rs

martinvonz:
The ssh_key_from_memory feature has had no effect in the last 6 years. It will be removed in the next major release (https://github.com/rust-lang/git2-rs/pull/1087) and it will thus be an error to request it.


sassman (approved):

sassman:
Thanks for your contribution!

Issue #1294, opened 2024-10-05
Implicit `.liquid` matching in `template.ignore` 


gibbz00:
I have a template repo with multiple files that have generated counterpart of the same name:

```
a.toml
README.md
README.md.liquid
cargo-generate.toml
```
```toml
# cargo-generate.toml
[template]
ignore = [
  "README.md"
]
````

Which results in the following after running `cargo-generate` (v. 0.22.0):

```
a.toml
.git
```

My expectation was that only `README.md.liquid` would remain. This time with its template fields expanded, and the `.liquid` extension stripped.

I may work around this using some Rhai post-hooks, but I would like to know if there's a better way to go about it.

#### Update

Having looked into the source code a bit, it seems as if extension stripping is done before template expansion, and that stripping should override any pre-existing file of the same name. Removing `README.md` from ignore does not seem to fix it though. The file is left as is, without the contents from the `.liquid` file.

#### Update 2

I tried to reproduce this in some integration tests, but they seem to pass without any problems, https://github.com/cargo-generate/cargo-generate/commit/5a45c7f646acfaf33d8e55e50a92846c1ab0c16f.

sassman:
That is an interesting case, I agree with your expectation of 

- the `README.md.liquid` should be used to
- replace the `README.md` 
- and **no** `README.md` must be on the ignore list

Do you have the repo already somewhere available?
Otherwise I might reuse your integration test.

gibbz00:
No sure why removing the respective files from the ignore list didn't seem fix things at first, but now it does 😅  

Could technically close this issue here and now. But maybe it would be better if it is closed together with the merging of a PR that includes the integration tests and some documentation that clarifies this behavior?

sassman:
Thanks for the PR that improves the clarity and tests.

PR #1295, opened 2024-10-05
Clarify `.liquid` overwrite and ignore behavior

gibbz00:
Fixes: #1294

### DIFF
--- a/guide/src/templates/ignoring.md
+++ b/guide/src/templates/ignoring.md
@@ -14,3 +14,6 @@ ignore = [
 ```
 
 Both files and folders may be ignored using this method, but currently wildcards are **not supported**.
+
+Note that `cargo-generate` checks for which files to ignore after the removal of any `.liquid` file extensions.
+Meaning; Setting `ignore` to `["file.txt"]` will result in the ignoring of a file named `file.txt.liquid`.
--- a/tests/integration/helpers/project_builder.rs
+++ b/tests/integration/helpers/project_builder.rs
@@ -1,6 +1,8 @@
+use assert_cmd::prelude::*;
 use indoc::indoc;
 use std::fs;
 use std::io::Write;
+use std::process::Command;
 use std::str;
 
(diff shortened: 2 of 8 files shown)


guide/src/templates/ignoring.md
```diff
@@ -14,3 +14,6 @@ ignore = [
 ```
 
 Both files and folders may be ignored using this method, but currently wildcards are **not supported**.
+
+Note that `cargo-generate` checks for which files to ignore after the removal of any `.liquid` file extensions.
```
sassman, line 18: I think this way brings more clarity.

tests/integration/helpers/project_builder.rs
```diff
@@ -95,6 +96,23 @@ impl ProjectBuilder {
             .success();
     }
 
+    fn git_commit(message: &str, current_directory: impl AsRef<std::path::Path>) {
+        Command::new("git")
+            .args([
+                "-c",
+                "user.name='Foo Bar'",
```
sassman, line 103: great idea, to inline the git user detail here, normally on CI we set this extra up, but doing it this way would make CI simpler and we can reduce more boiler plate 👍

sassman (approved):

tests/integration/helpers/project_builder.rs
gibbz00: Yeah, I was unsure how to run the tests locally too without adding this 😅 my git config picks user info depending if current directory is in my "personal" or "work" directory only.

guide/src/templates/ignoring.md
```diff
@@ -14,3 +14,6 @@ ignore = [
...
+Meaning; Setting `ignore` to `["file.txt"]` will result in the ignoring of a file named `file.txt.liquid`.
```
seffradev, line 19: I'd say "will result" could be clearer by saying "will also result". But it's just a technicality.

Issue #1340, opened 2024-12-23
Add new generated project as a new member to an existing workspace project


sassman:
Assuming I'm currently in a regular cargo workspace project with a few members like this:

```toml
[workspace]
members = ["day1", "day2", "day3", "day4", "day5", "day6"]
```

Further, assuming I have the template project in a folder `.template/day` at the project root, that is a regular cargo generate project with only 1 placeholder

When running now `cargo generate` to generate a new project, I want this new project being added to the workspace as a new member.

Ideally I don't need to specify any cli flag for this, the pure facts that my current working directory (`$CWD`) has a `Cargo.toml` that is a workspace manifest should be enough to trigger the addition to the members.

## Example

```
❯ cat Cargo.toml
[workspace]
members = ["day1", "day2", "day3", "day4", "day5", "day6"]

❯ cargo generate --path .template/day --name day7 --define day=7

❯ cat Cargo.toml
[workspace]
members = ["day1", "day2", "day3", "day4", "day5", "day6", "day7"]
```

PR #1342, opened 2024-12-23
feat: identify workspace project at parent folder and add as a member

sassman:
Solves #1340 

```sh
cargo generate --path .template/day --name day9 --define day=9
🔧   day: "9" (value from CLI)
🔧   Destination: /Users/sassman/workspace/AoC2024/day9 ...
🔧   project-name: day9 ...
🔧   Generating template ...
🔧   Moving generated files into: `/Users/sassman/workspace/AoC2024/day9`...
🔧   Initializing a fresh Git repository
✨   Done! New project created /Users/sassman/workspace/AoC2024/day9
⚠️   Project `day9` is already a member of the workspace
🔧   Project added as member to workspace `/Users/sassman/workspace/AoC2024/Cargo.toml`
```

## TODOs

- write integration tests for the new feature
- ensure that `git init` is not happening when part of a workspace project
- refactor a bit and introduce a context that does the environmental checks before so that 
  - the action plan like `git init` can be layed out up front
  - the user get's a chance to see what they will get upfront and opt out
  - e.g. interactively `Want to add this project as member to the workspace at: zyz? [Yn]`


James-Mart:
This seems to cause an issue in my case.
```
🔧   Moving generated files into: `.../services/user`...
Error: Failed to read .../services/user/Cargo.toml
```

(There is no Cargo.toml at `user/`, although there is at `services/`. But I don't want to add the generated project as a member to services/Cargo.toml)

PR #1373, opened 2025-01-24
feat(template): add multi select variables 

### LINKED ISSUE #1338: Feature request: allow multi select
I'm working on some template for myself, and I often have to work with different serde formats. What I would like is for there to be a multi select function. As an example in the `cargo-generate.toml` I'd imagine this something like: 

```toml
[placeholders.serde_formats]
type = "array"
prompt = "Which formats would you like?"
choices = ["JSON", "YAML", "CBOR", "TOML"] 
```
which I would then produce a prompt like:

```
🤷 Which formats would you like?
    [x] JSON
    [x] YAML
>   [ ]  CBOR
    [ ]  TOML
```

I'd want this to then produce an array containing the selected elements so that in the `Cargo.toml` on the template side I could do something like:

```toml
{% if "YAML" in serde_formats %}
serde_yaml = "0.9"
{% endif %}
{% if "JSON" in serde_formats %}
serde_json = "1.0.108"
{% endif %}
{% if "TOML" in serde_formats %}
toml_edit = "0.22.22"
{% endif %}
{% if "CBOR" in serde_formats %}
ciborium = "0.2.2"
{% endif %}
``` 

This is just an illustrative example, I don't actually care exactly what the formats or names turn out to be. Given what's already there, this seems like a doable feature, and if it's something the project is interested in I'd be open to making a PR for it.

savente93:
Fixes #1338 

should be noted that this is no where near done, as I'm still getting to grips with the code base, I'm just opening up prematurely so people can track and comment as necessary

### DIFF
--- a/src/interactive.rs
+++ b/src/interactive.rs
@@ -84,6 +84,7 @@ pub fn prompt_and_check_variable(
                 handle_string_input(provided_value, &variable.var_name, entry, &variable.prompt)
             }
         },
+        VarInfo::Array { entry } => todo!(),
     }
 }
 
@@ -95,6 +96,7 @@ pub fn variable(variable: &TemplateSlots, provided_value: Option<&impl ToString>
             Ok(Value::Scalar(as_bool.into()))
         }
         VarInfo::String { .. } => Ok(Value::Scalar(user_entry.into())),
+        VarInfo::Array { entry } => todo!(),
     }
 }
 
@@ -160,6 +162,45 @@ fn handle_string_input(
 }
 
 fn handle_choice_input(
+    provided_value: Option<String>,
+    var_name: &str,
+    choices: &Vec<String>,
+    entry: &StringEntry,
+    prompt: &Prompt,
+) -> Result<String> {
+    match provided_value {
+        Some(value) => {
+            if choices.contains(&value) {
+                Ok(value)
+            } else {
+                bail!(
+                    "{} {} \"{}\" {}",
+                    emoji::WARN,
+                    style("Sorry,").bold().red(),
+                    style(&value).bold().yellow(),
+                    style(format!("is not a valid value for {var_name}"))
+                        .bold()
+                        .red(),
+                )
+            }
+        }
+        None => {
+            let default = entry
+                .default
+                .as_ref()
+                .map_or(0, |default| choices.binary_search(default).unwrap_or(0));
+            let chosen = Select::with_theme(&ColorfulTheme::default())
+                .items(choices)
+                .with_prompt(&prompt.styled)
+                .default(default)
+                .interact()?;
+
+            Ok(choices.index(chosen).to_string())
+        }
+    }
+}
+
+fn handle_multi_select_input(
     provided_value: Option<String>,
     var_name: &str,
     choices: &Vec<String>,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -600,6 +600,7 @@ fn read_default_variable_value_from_template(slot: &TemplateSlots) -> Result<Str
             } => default.clone(),
             _ => return Err(()),
         },
+        VarInfo::Array { entry } => return Err(()),
         _ => return Err(()),
     };
     let (key, value) = (&slot.var_name, &default_value);
@@ -821,6 +822,7 @@ mod tests {
                     anyhow::bail!("Missing choices")
                 }
             }
+            VarInfo::Array { entry } => todo!(),
         })?
         .canonicalize()?;
         let expected = tmp.path().join("sub2").canonicalize()?;
@@ -878,6 +880,7 @@ mod tests {
                     anyhow::bail!("Missing choices")
                 }
             }
+            VarInfo::Array { entry } => todo!(),
         })?
         .canonicalize()?;
 
@@ -917,6 +920,7 @@ mod tests {
                     anyhow::bail!("Missing choices")
                 }
             }
+            VarInfo::Array { entry } => todo!(),
         })?
         .canonicalize()?;
         let expected = tmp.path().join("dir4").canonicalize()?;
--- a/src/project_variables.rs
+++ b/src/project_variables.rs
@@ -65,6 +65,7 @@ impl<T: AsRef<str>> From<T> for Prompt {
 /// Editor will never have choices
 #[derive(Debug, Clone)]
 pub enum VarInfo {
+    Array { entry: Box<Vec<String>> },
     Bool { default: Option<bool> },
     String { entry: Box<StringEntry> },
 }


sassman:
Awesome 👏 
If there is anything you need to know or would like to clarify feels free to ask questions directly on the PR with direct mentioning.

savente93:
@sassman so, one thing I am wondering as I am working on this, is how fancy we want to get with string parsing. Because things haved to be able to be specified from the ENV we will have to do some string parsing. like for example
```bash
MULTI_SELECT_CHOICES=a,c cargo generate ...
```
should logically end up in the template as the equivalent of `multi_select_choices = ['a','c']`, that's fine. However what about a silly malicious toy example like

```bash
ALLOWED_CSV_SEPS=,,;,|,\n cargo generate ...
```

i.e. would `,` be allowed to be a choice? or I guess whatever we choose as _our_ array seperator, is that allowed to be an option? obviously on the code side saying "no" is much easier, and I _think_ this is quite a niece case, but wanted to get your opinion. What do you think?

savente93:
This is all I'll have time for today, but I feel like I made some good progress. I added code to handle the prompting for multi selections and how to properly handle defaults in those cases. What I haven't figured out yet, is where I can define types like `multi-select` and `array` so that the template will accept them. @sassman could you point me in the right direction when you have a sec? thanks! I'll probably return to this sometime next week

src/interactive.rs
```diff
@@ -198,6 +202,80 @@ fn handle_choice_input(
     }
 }
 
+// simple function so we can easily get more complicated later if we need to
+fn parse_list(provided_value: &str) -> Vec<String> {
+    provided_value.split(',').map(|s| s.to_string()).collect()
```
sassman, line 207: for now I would put the choices separator `','` into a constant. 
So that it is easier to have it also configurable (via cli / env argument) in the future.

sassman:
> What I haven't figured out yet, is where I can define types like multi-select and array so that the template will accept them.

Maybe this function: https://github.com/cargo-generate/cargo-generate/blob/main/src/project_variables.rs#L443 helps a bit. 
The template in fact works only with strings, and we have an internal representation for a user input variable, but eventually all will be boiled down to a string that is injected into the context for a template rendering.

See also https://github.com/cargo-generate/cargo-generate/blob/main/src/lib.rs#L615 

Regarding the choices separator, it would be good to have it configurable, but having it pinned to a static or a const would also work for me. Not all niece case must be covered IMO.

savente93:
(summarising what I did for myself next time): managed to properly define the `array` and `multi-select` types. I'm now able to generate a very simple template with this now \0/ next port of call: making sure it ends up as an array on the liquid side, then testing after probably.

savente93:
so after some more investigation it seems that it is actually handling arrays correctly, I guess I just wasn't that familiar with how liquid handles these types of things. I also looked into the testing briefly, but I'm having trouble finding a good way to test this that doesn't basically just test the selection dependency. Got any ideas @sassman? aside from that I think we're basically done, so unless you have other requirements, I'll be opening this up for review soon after

sassman:
> I also looked into the testing briefly, but I'm having trouble finding a good way to test this that doesn't basically just test the selection dependency

Maybe this very test helps you: 
https://github.com/cargo-generate/cargo-generate/blob/main/tests/integration/template_config_file/placeholders.rs#L4

But in fact often it needs a whole setup with lots of combinations in order to verify a feature like this works.

Also, one more thing I forgot to mention so far, those variables are also made available to rhai hooks, you [can find a test case here](https://github.com/cargo-generate/cargo-generate/blob/main/tests/integration/hooks_and_rhai.rs), so probably you wanna look into this area too.

savente93:
@sassman I think we're ready for review \0/ thanks for the guidance, it's helped me keep motivated, in addition to answering my questions <3

src/interactive.rs
```diff
@@ -198,6 +209,91 @@ fn handle_choice_input(
     }
 }
 
+// simple function so we can easily get more complicated later if we need to
+fn parse_list(provided_value: &str) -> Vec<String> {
+    provided_value
+        .split(',')
```
sassman, line 215: Here the use of the separator const is missing.

src/lib.rs
```diff
@@ -611,6 +611,27 @@ fn read_default_variable_value_from_template(slot: &TemplateSlots) -> Result<Str
...
+        toml::Value::String(s) => Some(s.clone()),
+        toml::Value::Integer(s) => Some(s.to_string()),
+        toml::Value::Float(s) => Some(s.to_string()),
+        toml::Value::Boolean(s) => Some(s.to_string()),
+        toml::Value::Datetime(s) => Some(s.to_string()),
+        toml::Value::Array(s) => Some(
+            s.iter()
+                .filter_map(extract_toml_string)
+                .collect::<Vec<String>>()
+                .join(","),
```
sassman, line 629: Here same as above, better use the separator const

src/project_variables.rs
```diff
@@ -281,6 +290,16 @@ fn try_key_value_into_slot(
...
             }),
         },
+        SupportedVarType::Array => VarInfo::Array {
+            entry: Box::new(ArrayEntry {
+                default: if let Some(SupportedVarValue::Array(value)) = default_choice {
+                    Some(value)
+                } else {
+                    None
+                },
+                choices: choices.unwrap(),
```
sassman, line 300: Would be nice to avoid unwraps

src/project_variables.rs
```diff
@@ -425,6 +449,22 @@ fn extract_default(
...
+                .iter()
+                .filter(|f| !(f.is_table() && f.is_array()))
+                .map(|f| f.as_str().unwrap_or_default().to_string())
+                .collect();
+            if default_string_array.iter().all(|v| choices.contains(v)) {
+                Ok(Some(SupportedVarValue::Array(default_string_array.clone())))
+            } else {
+                Err(ConversionError::InvalidDefault {
+                    var_name: var_name.into(),
+                    default: default_string_array.join(","),
```
sassman, line 463: Also here the separator const should be used

tests/integration/hooks_and_rhai.rs
```diff
@@ -294,12 +298,17 @@ fn can_change_variables_from_pre_hook() {
...
+        .arg("-d")
+        .arg("multi=a,b")
         .assert()
         .success()
         .stdout(predicates::str::contains("Done!").from_utf8());
 
     assert!(dir.exists("script-project/PRE-TEST"));
-    assert!(dir.read("script-project/PRE-TEST").contains("bar"));
+    let pre_test = dir.read("script-project/PRE-TEST");
+    dbg!(&pre_test);
```
sassman, line 309: Pls remove the dbg

sassman (changes requested):
Found a couple of nits.
Overall nicely done 🚀

src/project_variables.rs
savente93: Good point, I used `unwrap_or_default` here which should give an empty `vec` which I think is fine, but let me know if you want a different solution.

sassman (approved):
thanks for your hard work, really appreciated 🚀

savente93:
Happy to help! It's been a fun one to work on. I can't merge the PR though :sweat_smile:

PR #1374, opened 2025-01-27
fix: inherit stdio for system::command calls

### LINKED ISSUE #1348: Feature request: inherit `system::command` stderr/stdout instead of capturing
I think this would improve situations where template authors are trying to debug `system::command` calls in their templates and where commands are long-running and users aren't sure if something is broken because there isn't any indication of progress.

CobaltCause:
Fixes #1348.

### DIFF
--- a/src/hooks/system_mod.rs
+++ b/src/hooks/system_mod.rs
@@ -73,18 +73,17 @@ fn run_command(
         return Err(format!("User denied execution of system command `{full_command}`.").into());
     }
 
-    let output = Command::new(name).args(args).output();
+    let status = Command::new(name).args(args).status();
 
-    match output {
-        Ok(output) => {
-            if output.status.success() {
+    match status {
+        Ok(status) => {
+            if status.success() {
                 Ok(Dynamic::UNIT)
             } else {
-                Err(format!(
-                    "System command `{full_command}` returned non-zero status: {}",
-                    output.status
+                Err(
+                    format!("System command `{full_command}` returned non-zero status: {status}")
+                        .into(),
                 )
-                .into())
             }
         }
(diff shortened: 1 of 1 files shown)


src/hooks/system_mod.rs
```diff
@@ -73,16 +73,16 @@ fn run_command(
...
-    match output {
-        Ok(output) => {
-            if output.status.success() {
+    match status {
+        Ok(status) => {
+            if status.success() {
                 Ok(Dynamic::UNIT)
             } else {
                 Err(format!(
                     "System command `{full_command}` returned non-zero status: {}",
```
sassman, line 84:
```suggestion
                    "System command `{full_command}` returned non-zero status: {status}",
```

src/hooks/system_mod.rs
```diff
@@ -73,16 +73,16 @@ fn run_command(
...
-                    output.status
+                    status
```
sassman, line 85:
```suggestion
```

src/hooks/system_mod.rs
CobaltCause: You might want to enable https://rust-lang.github.io/rust-clippy/master/#uninlined_format_args to automate this feedback

sassman (approved):

sassman:
just a note: during testing I figured that the outer error match always takes precedence. Yet the error message is there, so the goal is reached anyways.

```
        Err(e) => Err(format!("System command `{full_command}` failed to execute: {e}").into()),
```

The error looks then like:

```sh
foobarbak this one fails.. · yes
Error: ⛔   Failed executing script: post-script.rhai

Caused by:
    Runtime error: System command `foobarbak this one fails..` failed to execute: No such file or directory (os error 2)
```

PR #1376, opened 2025-01-27
feat(errors): fail on template errors

### LINKED ISSUE #1339: Feature request: strict/fail-fast mode
This is more of a feature request for the template authors than the users. I started making my own templates a few days ago, and had quite a few head scratches why my template files weren't being expanded at all. Turns out I made a syntax error, which I would have caught sooner if I had used the `--verbose` flag sooner which was on me. 

However, I do feel like this highlights two things I think a lot of folks like me could benefit from:
1. I think that template syntax error that prevent template expansion completely (at least within one a file) should raise a warning even in non-verbose mode. Even as a user trying to use a broken template I think being told about this could save a lot of head scratches. 
2. I'd appreciate a fail-fast or struck mode that halts execution on errors. Part of this is there in `--silent` which fails if not all variables are supplied, but there isn't anything similar for the liquid side as far as I can tell. Whether we should clean up after a failed run (i.e. remove output that was generated previously in a failed run) I don't actually care that much, you could argue for either. Happy to go with whatever maintainers think is best here. 

If this is something the project is interested in I'd be open to making a PR for this myself, but wanted to discuss it first.

savente93:
Fixes #1339 

So I'm aware we didn't agree on a design/solution for the linked issue before I started, so I'm happy to change the approach. My reasoning for the approach taken:

- I personally think that fatal errors such as a template containing syntax errors (since any non-trivial template is very unlikely to be a usable rust file) should not be hidden behind a `--verbose` flag (though admittedly I should have used that one when I ran into this issue myself, that's my bad).
- because expanded templates are unlikely to be valid rust, we should always error on invalid templates, but I am aware that this is a breaking change, and is also rather opinionated, so I can see not wanting this. 

So I've tried to strike a happy medium by:

- leaving `--verbose` alone, I wasn't  sure if it was used anywhere else
- introducing `--continue-on-error` and `--quiet` to maintain previous behaviour if users want to (though it is opt in)
- unless `--continue-on-error` is given, we exit unsuccessfully on a template syntax error, although we do still walk the entire tree, so all errors should get displayed
- I switched to reporting which files contained the errors from absolute to relative, since the absolute path is in a tmp dir for me which makes it harder to figure out which template file I need to go and edit. I think this is a nice UX improvement. 

This aligns things more with what I would have expected from the CLI options. One thing left to resolve would be that if this were merged we'd have `--quiet` and `--silent` which would be about quite different things, but I didn't touch that because I thought that was outside the scope of this PR. I'm happy to include any changes on this though if we can come up with better names.

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -95,8 +95,23 @@ pub struct GenerateArgs {
     pub force: bool,
 
     /// Enables more verbose output.
-    #[arg(long, short, action)]
+    #[arg(long, short, action, conflicts_with = "quiet")]
     pub verbose: bool,
+
+    /// Opposite of verbose, suppresses errors & warning in output
+    /// Conflicts with verbose, and requires the use of --continue-on-error
+    #[arg(
+        long,
+        short,
+        action,
+        conflicts_with = "verbose",
+        requires = "continue_on_error"
+    )]
+    pub quiet: bool,
+
+    /// Continue if errors in templates are encountered
+    #[arg(long, action)]
+    pub continue_on_error: bool,
 
     /// Pass template values through a file. Values should be in the format `key=value`, one per
     /// line
@@ -177,6 +192,8 @@ impl Default for GenerateArgs {
             name: None,
             force: false,
             verbose: false,
+            quiet: false,
+            continue_on_error: false,
             template_values_file: None,
             silent: false,
             config: None,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -512,7 +512,7 @@ fn expand_template(
         user_parsed_input.silent(),
         rhai_filter_files.clone(),
     );
-    template::walk_dir(
+    let result = template::walk_dir(
         &mut template_config,
         template_dir,
         &all_hook_files,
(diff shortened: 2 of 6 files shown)


src/args.rs
```diff
@@ -95,9 +95,24 @@ pub struct GenerateArgs {
...
+        short,
+        action,
+        conflicts_with = "verbose",
+        requires = "continue_on_error"
+    )]
+    pub quiet: bool,
+
+    /// Continue if errors in templates are encountered
+    #[arg(long, action)]
+    pub continue_on_error: bool,
```
sassman, line 114: I like this idea

sassman (commented):
Thanks for contributing.

I wonder since you started with `--test` and did not get the results you wished, maybe it is also good idea to automatically set max verbosity when testing. So that the errors would display in the case too.

WDYT?

savente93:
well that was surprisingly easy. I think those were all the changes that I had in mind for this PR :)

sassman (approved):

sassman:
tc

savente93:
The deployment seems to fail due to a missing dependency in the CI, so I don't think there's much I can do about that, but otherwise should all be fine now. Thanks for the review!

Issue #1444, opened 2025-04-03
the method `to_string` exists for enum `Option<&PackageName>`, but its trait bounds were not satisfied


davehorner:
**Describe the bug**
installing cargo-leptos got this
```  bash  Compiling cargo-generate v0.22.1
error[E0599]: the method `to_string` exists for enum `Option<&PackageName>`, but its trait bounds were not satisfied
    --> /Users/dhorner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cargo-generate-0.22.1/src/workspace_member.rs:110:62
     |
110  |         let name = manifest.package().unwrap().name.as_ref().to_string();
     |                                                              ^^^^^^^^^ method cannot be called on `Option<&PackageName>` due to unsatisfied trait bounds
     |
    ::: /Users/dhorner/.rustup/toolchains/nightly-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/option.rs:572:1
     |
572  | pub enum Option<T> {
     | ------------------ doesn't satisfy `_: Display` or `std::option::Option<&PackageName>: ToString`
     |
     = note: the following trait bounds were not satisfied:
             `std::option::Option<&PackageName>: std::fmt::Display`
             which is required by `std::option::Option<&PackageName>: ToString`
note: the method `to_string` exists on the type `&PackageName`
    --> /Users/dhorner/.rustup/toolchains/nightly-aarch64-apple-darwin/lib/rustlib/src/rust/library/alloc/src/string.rs:2747:5
     |
2747 |     fn to_string(&self) -> String;
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
help: consider using `Option::expect` to unwrap the `&PackageName` value, panicking if the value is an `Option::None`
     |
110  |         let name = manifest.package().unwrap().name.as_ref().expect("REASON").to_string();
     |                                                             +++++++++++++++++

For more information about this error, try `rustc --explain E0599`.
error: could not compile `cargo-generate` (lib) due to 1 previous error
error: failed to compile `cargo-leptos v0.2.32`, intermediate artifacts can be found at `/var/folders/nw/2x81h_b10qj6t5c2c90k_c300000gn/T/cargo-install49G4p6`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
```
did a cargo install cargo-generate; sample result
**make sure you**
- [ ] I'm on the latest rust version `rustup update`
[...]

PR #1443, opened 2025-04-03
fix: handle optional package.name field in workspace_member (see #1442)

lindestad:
### **Description**  
This pull request addresses [Issue #1442](https://github.com/OWNER/REPO/issues/1442), where the recent change in `cargo-util-schemas` made `package.name` optional. Previously, the code assumed `pkg.name.as_ref()` returned an `Option<...>`, causing a compile error. This update properly checks for the `[package]` section and uses `anyhow!` to bail out with a descriptive error if `package.name` is missing.

### **Key Changes**  
- Updated `WorkspaceMember::try_new` to safely unwrap `manifest.package()`.
- Replaced the old logic with a clearer `ok_or_else` pattern plus `anyhow!` for error creation.
- Ensured that if either `[package]` or `package.name` is missing, return a helpful error message instead of panicking or failing the compilation.

### **Testing**  
- Ran `cargo test` locally, confirming that all tests now pass on latest stable Rust.

PR #1445, opened 2025-04-03
fix(workspace_member): ensure package name is present in workspace ma…

davehorner:
…nifest

fix for https://github.com/cargo-generate/cargo-generate/issues/1444


sassman:
already fixed with #1443

Issue #1514, opened 2025-07-02
Excluded files are ignored when in a templated path


Gosu14:
**Description**
When an excluded file is contained in a templated path such as : `{{project-name}}/file_to_exclude.rs`, the file is completely ignored.

**To Reproduce**
1. `cargo generate --git https://github.com/Gosu14/cargo-generate-example`
2. Check the output. The file `excluded` has been ignored and is not in the output template.

**Expected behavior**
Excluded files should be skipped and not ignored. In that specific case of a file within a templated path, the file should be copied over the new path and skipped.

**Additional comments**
PR: #1515

sassman:
this is now part of https://github.com/cargo-generate/cargo-generate/releases/tag/v0.23.4

PR #1515, opened 2025-07-02
Fix: Excluded files in templated paths are now properly handled

Gosu14:
**Fix: Excluded files in templated paths are now properly handled**

Fix issue: #1514

**Problem**
Previously, when files were marked as excluded in the template configuration but were located within templated directory paths (e.g., `{{project-name}}/excluded-file`), the exclusion logic would simply skip processing them without considering that the directory path itself needed templating. This resulted in excluded files being ultimately ignored as they were not copied into the target directory.

**Objective**
Ensure that template exclusion rules work correctly regardless of whether the excluded files are in static or templated directory structures.

**Solution**
This fix enhances the exclusion handling logic in the `walk_dir` function to:

1. Apply filename substitution to excluded files: Even excluded files now go through the `substitute_filename` process to handle templated directory paths
2. Copy excluded files to templated locations: When an excluded file's path changes due to templating the file is copied to the new location
3. Clean up original files: After copying to the templated location, the original file in the templated path is removed
4. Preserve file content: Excluded files maintain their original content without any template processing

**Changes Made**

- Modified the `ShouldInclude::Exclude` branch in `walk_dir()` function in `template.rs`
- Added proper error handling for file operations during exclusion (I tried to follow codebase pattern and idiomatic Rust)
- Added test case `it_doesnt_process_excluded_files_in_templated_paths()` to verify the fix

**Test Coverage**
The new test verifies that:

- Excluded files in templated paths are moved to the correct final location
- Excluded files retain their original content (no template processing)
- Included files in the same templated paths are processed normally
- The exclusion configuration is respected

### DIFF
--- a/src/git/mod.rs
+++ b/src/git/mod.rs
@@ -74,7 +74,7 @@ pub fn remove_history(project_dir: &Path) -> io::Result<()> {
 
                 if e.to_string().contains("The process cannot access the file because it is being used by another process.") {
                     let wait_for = Duration::from_secs(2_u64.pow(attempt.sub(1).into()));
-                    warn!("Git history cleanup failed with a windows process blocking error. [Retry in {:?}]", wait_for);
+                    warn!("Git history cleanup failed with a windows process blocking error. [Retry in {wait_for:?}]");
                     sleep(wait_for);
                 } else {
                     return Err(e);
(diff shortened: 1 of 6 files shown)


sassman:
Thanks for the fix.

I will have a look at it in about 2 weeks, since I'm traveling at the moment.

Could you please have a look at the failing build and fix the lints as well?

sassman (approved):
Thanks for the PR, very much appreciated.

Issue #1501, opened 2025-06-07
test hooks::system_mod::tests::test_system_module fails if cat is not in /bin


Intbes:
### Issue

On Unix systems where system programs are not located in `/bin`, `test hooks::system_mod::tests::test_system_module` fails due to hardcoded location:

https://github.com/cargo-generate/cargo-generate/blob/17759b0a79c982fb538003070bfaf169595f4419/src/hooks/system_mod.rs#L199-L201

### Proposed fix

Replace `"/bin/cat"` with `"cat"`. In the same test, the command `pwd` is executed without specifying a hardcoded location and that doesn’t raise an issue:

https://github.com/cargo-generate/cargo-generate/blob/17759b0a79c982fb538003070bfaf169595f4419/src/hooks/system_mod.rs#L180

PR #1523, opened 2025-07-15
fix hardcoded path in tests

CobaltCause:
Fixes https://github.com/cargo-generate/cargo-generate/issues/1501


sassman (approved):
:rocket: thanks for your contribution

Issue #1516, opened 2025-07-02
Sometimes we _shouldn't_ add the new project to the containing workspace


dlaw:
Currently, if a new project is cargo-generated into a workspace, we _always_ add the project into the workspace's Cargo.toml.

But there are a few situations where we shouldn't do this:
1. If the new project's Cargo.toml already contains a `[workspace]` section (i.e. if we are using nested workspaces). This is common in embedded development.
2. If the workspace Cargo.toml contains an exclude pattern matching the new project, for example `exclude = ["*"]`.
3. Maybe the user just doesn't want us to rewrite the workspace Cargo.toml because that will reformat and reorder all of the toml.

Perhaps it is not worth the effort to write code for detecting cases 1 and 2. But, in any case, a `--no-workspace` command line flag would make it possible for the user to opt out of the undesired behavior.

sassman:
IMO a CLI flag like your proposed `--no-workspace` would be the first enabler to change the default behaviour. 

Going the extra mile to detect all the cases would be a nice to have. Only the check for the exclude pattern is IMO something that could be reasonably expected. 

Checking for for a nested workspaces could maybe lead to some unexpected surprises, but I have to admit I have never encountered such a structured project yet, so not sure if it is really worthwhile to spend effort on it.

dlaw:
Thanks for the comment. I agree, a `--no-workspace` flag would be sufficient and automatic detection may not be worth the effort. The nested workspace pattern is common in embedded projects, where there's an inner workspace for the embedded target and an outer workspace containing cross-platform libraries and APIs. Ferrous Systems has a nice blog post describing the rationale for this structure in more detail: https://ferrous-systems.com/blog/test-embedded-app/

Cheers!

sassman:
shipped with v0.23.6

PR #1531, opened 2025-08-11
feat: add --no-workspace flag that skips workspace addition

iainmcgin:
This change implements the suggested `--no-workspace` flag from the discussion in #1516. When specified, no attempt is made to try and find or modify a parent `Cargo.toml` workspace members list to include the newly generated project.


sassman:
Thanks for your feature, we took this further by merging #1570

PR #1532, opened 2025-08-12
fix: remove unnecessary use of concat!

iainmcgin:
Spotted this in the CI lint logs for
`cargo clippy --all-targets --all-features -- -D warnings`.


sassman:
the issue was meanwhile fixed.

PR #1556, opened 2025-09-20
Fix stale link text

Athosvk:
_Please note: I've read the contributing guidelines about linking this to an issue, but this change was so trivial that I think it's warranted for it to just be a PR alone. Let me know if I should change that._

Noticed this upon running `cargo generate` for the first time, as it was pointing to a stale resource :)


sassman (approved):
:rocket: thanks for your contribution

sassman (approved):
:rocket: thanks for your contribution

PR #1622, opened 2026-03-18
feat: add opt-in ui-next feature for the cliclack terminal UI

sassman:
## Summary

The full cliclack UI now ships behind a `--features=ui-next` opt-in instead of replacing the default. Existing users see no change; anyone who wants to try the new look flips a build flag.

## What ships

- **`ui-next` feature toggle** — `cliclack` + `fastrand` become optional deps; classic dialoguer/indicatif stays the default.
- **`ui::` facade** — shared code (`lib.rs`, `hooks`, `favorites`, `project_variables`, …) routes status output through one API so either backend renders in-flow.
- **Parallel modules** — `src/interactive/{classic,next}.rs` and `src/favorites/{classic,next}.rs` are cfg-selected at compile time; identical public surface.
- **Deprecation** — classic UI public fns carry `#[deprecated(since = "0.24.0", note = "…")]` so downstream users see the signal in rustdoc / build output.
- **Tip banner** — one-line 💡 hint at startup pointing at `cargo install --features=ui-next --locked cargo-generate`; suppressed under `--silent`, `--list-favorites`, `CARGO_GENERATE_NO_TIP`, `NO_COLOR`, or non-tty stderr.
- **Icon consolidation** — one 🔸 for info and warning (colour still differentiates); the old `⚠️` often rendered as plain text because of its variation selector.

## Try the preview

```sh
cargo install --features=ui-next --locked cargo-generate
```

## Test plan

- [x] `cargo test --lib` — 171 passing (classic) / 189 passing (ui-next)
- [x] `cargo test --test integration` — 115/115 on both feature sets
- [x] `cargo clippy --locked --all-targets -- -D warnings` clean on both
- [x] Manual `cargo generate` — classic path terse, no box drift, tip renders
- [x] Manual `cargo generate --features=ui-next` — cliclack box intact end-to-end
- [x] `--silent` skips the tip; classic favorites list renders through stderr

## What ui-next looks like

- success case

Pasted image 20260319002930

- error case

image


ohmsmithtrip:
This migration sounds promising! The cohesive interactive experience will greatly improve usability. It would be great to see examples of how the new prompts compare visually and functionally to the previous implementation. Looking forward to the updates!

PR #1653, opened 2026-04-20
Fix typos in guide

c-git:
# Update deprecation note with correct link

Seems this may have been on the same page before but was later moved. Didn't check the history but I think this is where the link was meant to point to.


sassman:
thank a lot for the eagle eye!

sassman (approved):

c-git:
Sorry found another one. Going to put a reminder to come back in a few days to mark it ready again. Don't want to waste your time and hopefully a reminder means I won't forget to come back.

sassman:
looks all good, should we make the PR final then?

c-git:
Yeah I think so. I'm done with the template so not actively digging through the docs anymore so unlikely to find any more in the near future. Thanks for everything.

PR #1654, opened 2026-04-20
Fix typos in docs

c-git:
Found another one will keep this as a draft until I'm done reading in case I find more.


c-git:
Turns out I was almost at the end.

PS. There were other options to fix the sentence but I chose what felt like the simplest option.

For example instead of just deleting `now` as I did it seemed to read fine as

```
The answer to this question is a vital build pipeline for your template project. This has challenge now gotten
much simpler to solve with the new official...
```

But that felt like it wouldn't age well.

c-git:
I just realized https://github.com/cargo-generate/cargo-generate/pull/1653 is still open I will add this to that PR instead.

Issue #1687, opened 2026-06-12
Conflicting implementations of trait in crate time, Rust 1.95


Vanuan:
**Describe the bug**
Running `cargo install cargo-generate` throws compilation errors

**To Reproduce**
cargo install cargo-generate
```
   Compiling cargo-generate v0.23.11
error[E0119]: conflicting implementations of trait `From<format_description::parse::format_item::HourBase>` for type `<format_description::parse::format_item::HourBase as format_description::parse::format_item::ModifierValue>::Type`
  --> /home/john/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cargo-generate-0.23.11/src/project_variables.rs:59:1
   |
59 | impl<T: AsRef<str>> From<T> for Prompt {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: conflicting implementation in crate `time`:
           - impl From<format_description::parse::format_item::HourBase> for <format_description::parse::format_item::HourBase as format_description::parse::format_item::ModifierValue>::Type;
   = note: upstream crates may add a new impl of trait `std::convert::AsRef<str>` for type `time::format_description::parse::format_item::HourBase` in future versions

error[E0119]: conflicting implementations of trait `From<format_description::parse::format_item::HourBase>` for type `<format_description::parse::format_item::HourBase as format_description::parse::format_item::ModifierValue>::Type`
   --> /home/john/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cargo-generate-0.23.11/src/user_parsed_input.rs:456:1
    |
456 | / impl<T> From<T> for TemplateLocation
457 | | where
458 | |     T: AsRef<Path>,
    | |___________________^
    |
    = note: conflicting implementation in crate `time`:
            - impl From<format_description::parse::format_item::HourBase> for <format_description::parse::format_item::HourBase as format_description::parse::format_item::ModifierValue>::Type;
```

**Expected behavior**
Expect `cargo generate` command available

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Version: 26.04
 - Build tools: gcc (Ubuntu 15.2.0-16ubuntu1) 15.2.0
 - Shell: bash
 - `rustc --version`: 1.95.0 (59807616e 2026-04-14)
 - `cargo generate --version`: N/A

**Additional context**

What am I doing wrong?

Vanuan:
Workarounds I tried:
[...]

PR #1688, opened 2026-06-12
fix: resolve conflicting From trait implementations for Prompt and TemplateLocation

Hank1212121:
The recent Rust compiler has stricter trait coherence and orphan rules, which caused blanket generic implementations of `From<T>` for `Prompt` and `TemplateLocation` to conflict with upstream crates like `time`.

This PR fixes the issue by replacing the blanket generic implementations with concrete implementations for common types (`&str`, `String`, `&Path`, `PathBuf`, etc.).

### DIFF
--- a/src/project_variables.rs
+++ b/src/project_variables.rs
@@ -56,9 +56,21 @@ impl Prompt {
     }
 }
 
-impl<T: AsRef<str>> From<T> for Prompt {
-    fn from(value: T) -> Self {
-        Self::new(value.as_ref(), None)
+impl From<&str> for Prompt {
+    fn from(value: &str) -> Self {
+        Self::new(value, None)
+    }
+}
+
+impl From<String> for Prompt {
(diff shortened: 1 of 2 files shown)


sassman (approved):

sassman:
Thanks a lot

Vanuan:
Oh, so I'm not alone. Apparently `time` crate updated today that broke `cargo generate`. `--locked` flag worked, so the workflow root cause so to speak is that 0.x dependencies can break any time if not specified in Cargo.toml

This should fix #1687
More robust fix would be to lock all `0.x` dependencies in Cargo.toml and update only as needed. Alternatively, README should recommend to use `cargo install --locked`.

The default behaviour of `cargo install`, to ignore the lockfile and only look at Cargo.toml. The `--locked` flag forces to to use the lockfile included in the package.

This reminds me early days of bundler/ruby/Gemfile.lock where first few year it was kinda useless.

PR #1714, opened 2026-07-08
fix: use array defaults in silent mode

ychampion:
## Summary

- Use configured array placeholder defaults when running in `--silent` mode.
- Add an integration test covering silent-mode generation without a provided array value.

## Why

Fixes #1621. Array defaults were parsed from `cargo-generate.toml`, but silent-mode default lookup only accepted bool and string placeholders before prompting/validation.

## Validation

- `cargo test it_uses_array_defaults_in_silent_mode --test integration -- --nocapture` — passed
- `cargo test template_config_file::placeholders --test integration` — passed
- `cargo test` — passed
- `cargo run -- generate --silent --path ./example-templates/arrays --name test-array-default --define description='this is a test'` — failed before the fix and passed after the fix

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -632,6 +632,10 @@ fn read_default_variable_value_from_template(slot: &TemplateSlots) -> Result<Str
             } => default.clone(),
             _ => return Err(()),
         },
+        VarInfo::Array { entry } => match &entry.default {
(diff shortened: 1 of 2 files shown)


sassman (approved):
thanks a lot for this fix

PR #1716, opened 2026-07-09
fix: support array placeholders in conditionals

### LINKED ISSUE #1429: Conditionals for empty arrays?
I am not sure if this is already supported in some way but I couldn't find a way to make it work but it would be useful if it was possible to ignore additional files if an array placeholder is empty (e.g. in my case a script to push deploy branches for various platforms but I select 0 platforms for that placeholder).

If this is already supported it would be nice if it was documented somewhere or part of one of the included examples since it is not obvious which syntax the conditional expressions actually use (that in itself should probably be documented somewhere too).

ychampion:
## Summary

- Expose placeholder arrays as Rhai arrays when evaluating `conditional.'...'` blocks.
- Add regression coverage for `.contains(...)` and `.is_empty` array conditionals.
- Document that conditional expressions use Rhai syntax.

## Why

Fixes #1429. Array placeholders were already present in the template values, but the conditional evaluator only returned scalar values to Rhai, so array conditions were treated as missing/false.

## Validation

- `cargo test --test integration array_ -- --nocapture` — failed before the fix and passed after
- `cargo test --test integration conditionals:: -- --nocapture` — passed
- `cargo test --lib hooks::variable_mod -- --nocapture` — passed
- `cargo test` — passed
- `git diff --check` — passed

### DIFF
--- a/guide/src/templates/conditional.md
+++ b/guide/src/templates/conditional.md
@@ -4,6 +4,8 @@ Using `cargo-generate.toml`, values and some [`Rhai`] syntax, the template autho
 
 `include`, `exclude`, `ignore` and `placeholders` can all be used in sections that are only used based upon the value of one or more values, possibly input by the user using the interactive prompt (if the values in question are defined as placeholders in the non-conditional section).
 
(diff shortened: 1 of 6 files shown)


sassman (approved):
🙏🙏🙏

PR #1718, opened 2026-07-09
fix: use default sub-template in silent mode

### LINKED ISSUE #1181: Default sub template is not automatically used in silent mode (without user interaction)
**Describe the bug**

https://cargo-generate.github.io/cargo-generate/usage.html#templates-in-subfolders:

> Multiple sub-templates can also be configured in the `cargo-generate.toml` file like this:
>
> ```toml
> [template]
> sub_templates = ["folder1", "folder2"]
> ```
>
> Doing so also sets the order when `cargo-generate` asks what to expand, while *the first option will be the default*.

Running CLI in silent mode it will still ask for the sub template even if the first should be selected as a default one.

**make sure you**
- [x] I'm on the latest rust version `rustup update`
- [x] I'm on the latest cargo-generate version `cargo install cargo-generate`
- [ ] In case you have an own `.gitconfig`, provide all sections that contains one of: `insteadOf`, `url`

**To Reproduce**
1. `cargo-generate generate -s -n name dioxuslabs/dioxus-template`
2. output it generated..
  ```
  ⚠   Favorite `dioxuslabs/dioxus-template` not found in config, using it as a git repository: https://github.com/dioxuslabs/dioxus-template.git
  ? 🤷   Which sub-template should be expanded? ›
  ❯ Web
    Liveview
    Fullstack
    Desktop
    TUI
  ```
3. See ~~error~~ user interaction

**Expected behavior**
No user interaction, default sub template should be used.

 - OS: Pop!_OS
 - Version: 22.04
 - Build tools: gnu
 - Shell: zsh
 - `rustc --version`: 1.80.0-nightly (1fffb2a35 2024-04-28)
 - `cargo generate --version`: 0.20.0

ychampion:
## Summary

- use the configured default sub-template instead of prompting in `--silent` mode
- add unit and CLI regression coverage for the default sub-template path
- update existing formatting-only Clippy warnings from Rust 1.97

## Why

Fixes #1181. `sub_templates` already mark the first entry as the default, so silent mode should use it rather than waiting for an interactive selection.

## Validation

- `cargo test --lib select_sub_template -- --nocapture` — passed
- `cargo test --lib remove_dir_files_removes_existing_files_and_directories -- --nocapture` — passed
- `cargo test --lib resolve_template_dir_uses_default_subtemplate_in_silent_mode -- --nocapture` — passed
- `cargo test --test integration it_uses_the_default_subtemplate_in_silent_mode -- --nocapture` — passed
- `cargo test --lib auto_locate_template -- --nocapture` — passed
- `cargo test` — passed
- `git diff --check` — passed

### DIFF
--- a/src/git/clone_tool.rs
+++ b/src/git/clone_tool.rs
@@ -67,10 +67,7 @@ impl<'cb> RepoCloneBuilder<'cb> {
             self.gitconfig = Some(Config::open(gitconfig.as_path())?);
 
             if let Some(url) = gitconfig::resolve_instead_url(&self.url, gitconfig)? {
-                debug!(
-                    "{} gitconfig 'insteadOf' lead to this url: {}",
-                    &WRENCH, url
-                );
+                debug!("{} gitconfig 'insteadOf' lead to this url: {}", WRENCH, url);
                 self.url = url;
             }
         }
--- a/src/ignore_me.rs
+++ b/src/ignore_me.rs
@@ -101,3 +101,24 @@ pub fn remove_dir_files(files: impl IntoIterator<Item = impl Into<PathBuf>>, ver
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::remove_dir_files;
+    use std::fs;
+
+    #[test]
+    fn remove_dir_files_removes_existing_files_and_directories() {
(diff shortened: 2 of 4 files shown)


sassman (approved):
🙏

PR #1747, opened 2026-08-18
feat: support value/label choices for placeholders

### LINKED ISSUE #1698: feat: Choices with label and value
Wouldn't it be usefull to separate choices labels specified in `cargo-generate.toml` and ids used in liquid templates. This would make it easier to modify the choice without combing through all the liquid templates.

ChrisJr404:
Closes #1698.

## What

Lets template authors give a `choices` entry a display **label** that differs from the **value** handed to the template, without any workaround (previously the value shown to the user and the value substituted into the template were always the same string, so a hint like `1.3.7 (recommended)` had to be stripped back off in a rhai pre-hook).

A choice can now be written either as a plain string (exactly as before) or as a table with an explicit `value` and an optional `label`:

```toml
[placeholders.version]
type = "string"
prompt = "Which version?"
choices = [
    { value = "recommended",  label = "1.3.7 (recommended)"  },
    { value = "experimental", label = "1.4.3 (experimental)" },
    "older",
]
default = "recommended"
```

The prompt shows the labels, while `{{ version }}` (and `default`, `--define` and `regex`) always see the `value`:

```liquid
{% if version == "recommended" %}framework = { version = "1.3.7" }{% endif %}
```

This matches the `{ value, label }` shape agreed in the issue.

## Notes

- **Fully backwards compatible** — a plain string is treated as `value == label`, and the two forms can be mixed within one array. Existing templates are unchanged.
- Works for both `string` (single-select) and `array` (multi-select) placeholders.
- A table without a `value` (or with a non-string `value`/`label`) is rejected with a clear error.
- Regex validation and `default` matching are checked against the `value`, never the label.
- Internally, choices are now carried as a small `Choice { value, label }` type; the two previously-duplicated choice-parsing arms are unified into one helper.

## Tests

- New unit tests in `project_variables` covering: value/label tables, mixing with plain strings, label defaulting to value, missing-`value` error, regex checked against value not label, multi-select tables, and `default` matched against value.
- New integration tests: value used (not label) in silent mode, value/label choices for arrays via `--define`, and a label supplied as a value being rejected.
- Full suite green: 188 lib + 116 integration tests, plus `cargo fmt --check` and `cargo clippy --all-targets` clean.

I also documented the new form in the guide (`template_defined_placeholders.md`). Happy to adjust naming or error wording to taste.


tests/integration/template_config_file/placeholders.rs
```diff
@@ -259,3 +259,118 @@ fn it_renders_arrays_as_list() {
...
+                    { value = "logging", label = "Logging (tracing)" },
+                    "async",
+                ]
+            "#},
+        )
+        .file(
+            "features_as_list",
+            indoc! {r#"
+                [{%- for f in features -%}
+                    "{{ f }}"{% unless forloop.last %}, {% endunless -%}
```
sassman, line 322: that `{{ f }}` by default would yield the value is useful.

But I can imagine cases where a template author might want the control to render also `{{ f.label }}`

Lets explore if we can have both at the same time?

sassman (approved):

tests/integration/template_config_file/placeholders.rs
sassman: I will check that in a follow up PR

sassman:
Thanks a lot for your contribution, a very nice feature 🙏

